PR #6, opened 2024-10-17
MODLD-535: Disable automatic transformation of MARC bibliographic rec…

askhat-abishev:
…ords

### DIFF
--- a/src/test/java/org/folio/linked/data/integration/kafka/listener/SourceRecordDomainEventListenerIT.java
+++ b/src/test/java/org/folio/linked/data/integration/kafka/listener/SourceRecordDomainEventListenerIT.java
@@ -1,24 +1,21 @@
 package org.folio.linked.data.integration.kafka.listener;
 
 import static org.folio.linked.data.domain.dto.SourceRecordDomainEvent.EventTypeEnum.CREATED;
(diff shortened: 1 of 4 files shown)


src/test/java/org/folio/linked/data/integration/kafka/listener/SourceRecordDomainEventListenerIT.java
```diff
@@ -43,31 +40,39 @@ static void setup(@Autowired KafkaAdminService kafkaAdminService) {
   }
 
   @Test
-  void shouldConsumeSrsDomainEvent() {
+  void shouldNotHandleSrsDomainEvent_whenSourceRecordType_isMarcBib() {
```
PBobylev, line 43: Let's not delete tests, we will switch this feature on later and we know it for sure. Let's just disable it.

PBobylev (changes requested):

src/test/java/org/folio/linked/data/integration/kafka/listener/SourceRecordDomainEventListenerIT.java
askhat-abishev: done

PR #7, opened 2024-10-18
MODLD-535: Disable automatic transformation of MARC bibliographic records

askhat-abishev:


AndreiBordak (approved):

PBobylev (approved):

pkjacob:
Hi @askhat-abishev, As I mentioned in the previous PR, there should be a log entry when the importMarcRecord call is successful. I do not see any logs in this case. 

Something like `MARC BIB record with SRS ID {} is successfully imported to graph resource with ID {}`

Also, can you please check if we have enough logs for the `preview` API? Something like `Returning preview for SRS ID {}`

src/test/java/org/folio/linked/data/integration/kafka/listener/SourceRecordDomainEventListenerIT.java
```diff
@@ -42,6 +45,7 @@ static void setup(@Autowired KafkaAdminService kafkaAdminService) {
     kafkaAdminService.createTopics(TENANT_ID);
   }
 
+  @Disabled
```
pkjacob, line 48: There are sonar violations for these disabled tests.

askhat-abishev: fixed

pkjacob (approved):

PR #8, opened 2024-10-21
MODLD-504: fix authority update and cover with special IT (#4)

PBobylev:


AndreiBordak (approved):

src/test/java/org/folio/linked/data/test/TestUtil.java
```diff
@@ -125,4 +130,28 @@ public static void awaitAndAssert(ThrowingRunnable throwingRunnable) {
...
+
+  public static void assertAuthority(Resource resource,
+                               String label,
+                               boolean isActive,
+                               boolean isPreferred,
+                               Resource replacedBy) {
+    assertThat(resource)
+      .hasFieldOrPropertyWithValue("label", label)
+      .hasFieldOrPropertyWithValue("active", isActive)
+      .satisfies(r -> assertThat(r.getDoc()).isNotEmpty())
```
askhat-abishev, line 142: minor
looks like redundant

askhat-abishev (approved):

src/test/java/org/folio/linked/data/test/TestUtil.java
PBobylev: but if we omit it, next line will throw NPE in case r.getDoc() is null, right? So let's keep it

src/test/java/org/folio/linked/data/test/TestUtil.java
```diff
@@ -125,4 +130,28 @@ public static void awaitAndAssert(ThrowingRunnable throwingRunnable) {
...
+      .satisfies(r ->
+        assertThat(resource.getDoc().get(RESOURCE_PREFERRED.getValue()).get(0).asBoolean()).isEqualTo(isPreferred)
+      )
+      .satisfies(r -> assertThat(r.getOutgoingEdges()).isNotEmpty())
```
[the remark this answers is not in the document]
PBobylev: but if we omit it, next line will throw NPE in case r.getOutgoingEdges() is null, right? So let's keep it

PR #22, opened 2024-10-30
MODLD-563: API for loading record to LDE

askhat-abishev:


src/main/java/org/folio/linked/data/service/resource/ResourceMarcServiceImpl.java
```diff
@@ -98,17 +100,25 @@ public Boolean isSupportedByInventoryId(String inventoryId) {
...
+  public ResourceIdDto importMarcRecord(String inventoryId) {
+    return getResource(inventoryId)
+      .map(resource -> {
+        resource.getFolioMetadata().setSource(LINKED_DATA);
+        return resource;
+      })
+      .map(this::saveMarcResource)
+      .map(String::valueOf)
+      .map(id -> new ResourceIdDto().id(id))
+      .orElse(null);
```
AndreiBordak, line 119: i'd suggest to return empty Optional here instead of null. ResponseEntity should work fine with Optional objects

PBobylev (changes requested):

src/main/java/org/folio/linked/data/service/resource/ResourceMarcServiceImpl.java
askhat-abishev: done

AndreiBordak: ResponseEntity<Optional<ResourceIdDto>> can automatically solve this issue as per the documentation:
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/ResponseEntity.html#of(java.util.Optional)

it will lead to 404 NOT_FOUND in case of Optional.empty(), so we don't need to throw an exception unless we want to customize handling

src/main/java/org/folio/linked/data/service/resource/ResourceMarcServiceImpl.java
```diff
@@ -87,26 +92,34 @@ public ResourceMarcViewDto getResourceMarcView(Long id) {
...
+  }
+
+  @Override
+  public ResourceIdDto importMarcRecord(String inventoryId) {
+    return getResource(inventoryId)
+      .map(resource -> {
+        resource.getFolioMetadata().setSource(LINKED_DATA);
+        return resource;
+      })
+      .map(this::saveMarcResource)
```
pkjacob, line 119: Hi @askhat-abishev,

The current implementation of `saveMarcResource` update/replace the existing resource if resource already exists in the system. However, for this use case, we need to throw an error when the resource already exists in the system. Please see the comment in https://folio-org.atlassian.net/browse/UILD-408

So, we need to call a different version of `saveMarcResource` here with the following logic

```
    if (folioMetadataRepository.existsBySrsId(modelResource.getFolioMetadata().getSrsId())) {
      throw new AlreadyExistsException("MARC record having srsID {} is already imported to data graph");
    }
    
    if (resourceRepo.existsById(modelResource.getId()) || ) {
      throw new AlreadyExistsException("Another resource with ID {} already exists in the graph");
    }
    
    return createResource(mapped);
```

src/main/java/org/folio/linked/data/service/resource/ResourceMarcServiceImpl.java
askhat-abishev: @AndreiBordak, I don't see a way to specify Optional using swagger

src/main/java/org/folio/linked/data/service/resource/ResourceMarcServiceImpl.java
askhat-abishev: done

AndreiBordak (approved):

PBobylev (approved):

src/main/java/org/folio/linked/data/service/resource/ResourceMarcServiceImpl.java
```diff
@@ -273,7 +273,7 @@ private Long save(org.folio.ld.dictionary.model.Resource modelResource) {
       throw new AlreadyExistsException(format("MARC record having srsID %s is already imported to data graph", srsId));
     }
 
-    return createResource(resourceModelMapper.toEntity(modelResource));
+    return saveAndPublishEvent(resourceModelMapper.toEntity(modelResource), ResourceUpdatedEvent::new);
```
askhat-abishev, line 276: while working on karate tests I stumbled upon such check
image

therefore I've changed ResourceCreatedEvent to ResourceUpdatedEvent

pkjacob: We have to publish `ResourceCreatedEvent` @askhat-abishev. 
Publishing `ResourceUpdatedEvent` here will be incorrect. Let me try to check the problem today.

src/main/java/org/folio/linked/data/service/resource/ResourceMarcServiceImpl.java
```diff
@@ -230,8 +245,43 @@ private boolean isMonograph(String leader) {
...
+  private Optional<org.folio.ld.dictionary.model.Resource> getResource(String inventoryId) {
+    return getRecord(inventoryId)
+      .map(HttpEntity::getBody)
+      .map(Record::getParsedRecord)
+      .map(ParsedRecord::getContent)
+      .map(this::toJsonString)
+      .flatMap(marcBib2ldMapper::fromMarcJson);
+  }
+
+  private Long save(org.folio.ld.dictionary.model.Resource modelResource) {
```
pkjacob, line 265: These changes will work. However, the class has become complex and less readable due to different `save**` and `create**` methods. I think we can split this class into two

1. ResourceMarcAuthorityService -> class for Marc authority records
2. ResourceMarcBibService -> class for Marc bib records.

I have created a DRAFT PR https://github.com/folio-org/mod-linked-data/pull/25 with this idea. Can you please take a look? I think we can implement this in scope of MODLD-535

@PBobylev  @AndreiBordak, What do you think?

src/main/java/org/folio/linked/data/service/resource/ResourceMarcServiceImpl.java
```diff
@@ -230,8 +245,43 @@ private boolean isMonograph(String leader) {
...
+    if (resourceRepo.existsById(id)) {
+      throw new AlreadyExistsException(format("Another resource with ID %s already exists in the graph", id));
+    }
+
+    var srsId = modelResource.getFolioMetadata().getSrsId();
+    if (folioMetadataRepository.existsBySrsId(srsId)) {
+      throw new AlreadyExistsException(format("MARC record having srsID %s is already imported to data graph", srsId));
+    }
+
+    return saveAndPublishEvent(resourceModelMapper.toEntity(modelResource), ResourceUpdatedEvent::new);
```
pkjacob, line 276: Hi @askhat-abishev,
Can we add a comment here why we are emitting ResourceUpdatedEvent?

Emitting ResourceCreatedEvent here will result in sending a "CREATE_INSTANCE" event to Inventory and hence will result in creating a duplicate event in inventory. By emitting ResourceUpdatedEvent, we will send an "UPDATE_INSTANCE" event to inventory there by switching the source of the record as "LINKED_DATA" in inventory.

src/main/java/org/folio/linked/data/service/resource/ResourceMarcServiceImpl.java
pkjacob: Got it Askhat. Let us update the karate test to ensure that the source of the record is switched from MARC to linked-data after the import.

src/main/java/org/folio/linked/data/service/resource/ResourceMarcServiceImpl.java
```diff
@@ -230,8 +245,43 @@ private boolean isMonograph(String leader) {
...
+    var id = modelResource.getId();
+    if (resourceRepo.existsById(id)) {
+      throw new AlreadyExistsException(format("Another resource with ID %s already exists in the graph", id));
+    }
+
+    var srsId = modelResource.getFolioMetadata().getSrsId();
+    if (folioMetadataRepository.existsBySrsId(srsId)) {
+      throw new AlreadyExistsException(format("MARC record having srsID %s is already imported to data graph", srsId));
+    }
+
```
pkjacob, line 275: Hi @askhat-abishev, I do not think anything will get logged when the API call is made. Can you please ensure that log messages are added?

pkjacob (approved):

src/main/java/org/folio/linked/data/service/resource/ResourceMarcServiceImpl.java
AndreiBordak: @pkjacob looks like a good idea

PR #25, opened 2024-10-31
cleanup

pkjacob:


src/main/java/org/folio/linked/data/integration/kafka/listener/handler/SourceRecordDomainEventHandler.java
```diff
@@ -73,43 +68,19 @@ private boolean notProcessableEvent(SourceRecordDomainEvent event, SourceRecordT
     return false;
   }
 
-  private void saveBib(SourceRecordDomainEvent event) {
```
pkjacob, line 76: Remove these methods in scope of MODLD-535.
We will not do automatic import of Marc BIB record at least till 2026. So, no point in retaining these unnecessary methods for future.

PR #36, opened 2024-11-07
Modld-568: Fix Module evaluation findings for TC compliance

pkjacob:


descriptors/ModuleDescriptor-template.json
```diff
@@ -161,6 +161,12 @@
     }
   ],
   "requires": [],
+  "optional": [
+    {
+      "id": "source-storage-records",
```
pkjacob, line 166: This is needed as we make call to `GET /source-storage/records/{inventoryId}/formatted`

askhat-abishev: maybe we should place it into ```requires``` instead?

askhat-abishev (approved):

descriptors/ModuleDescriptor-template.json
pkjacob: Done

PR #43, opened 2024-11-14
MODLD-583: LCCN Validation of Pattern

askhat-abishev:


src/main/java/org/folio/linked/data/configuration/ValidatorsConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+@Configuration
+public class ValidatorsConfiguration {
+
+  @Bean(VALIDATOR_CHAIN_RESOURCE_SAVE)
+  public LdValidator<ResourceRequestDto> resourceSaveValidatorsChain(LdValidator<ResourceRequestDto> lccnValidator) {
+    // in future task we can filter out disabled validators here
+    return lccnValidator;
+  }
+
+  @Bean(VALIDATOR_CHAIN_RESOURCE_UPDATE)
```
pkjacob, line 20: I think we do not need two separate validator chains for save an update now. May be we can start with one validator (for both save and update) and create separate validator beans if a need arise in future.

pkjacob:
Hi @PBobylev, @askhat-abishev, @AndreiBordak,
In order to make the DTO validation approach consistent, I think it will be good to refactor the current `PrimaryTitleDtoValidator` to follow the approach in this PR. What do you think?

src/main/java/org/folio/linked/data/configuration/ValidatorsConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+import org.folio.linked.data.service.validation.LdValidator;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class ValidatorsConfiguration {
+
+  @Bean(VALIDATOR_CHAIN_RESOURCE_SAVE)
+  public LdValidator<ResourceRequestDto> resourceSaveValidatorsChain(LdValidator<ResourceRequestDto> lccnValidator) {
+    // in future task we can filter out disabled validators here
```
pkjacob, line 16: One scenario to think about: Let us say, during the application startup the APIs for checking the enable/disable flag is not available. In that case how will the application behave?

AndreiBordak: agree with PK, also we can automatically build a chain of validators by passing List<LdValidator<T>> validators and combine them using Stream#reduce()

src/main/java/org/folio/linked/data/configuration/ValidatorsConfiguration.java
AndreiBordak: we'll probably need separate validators for LCCN duplication validation, we should validate LCCN for duplicate only on create right?

src/main/java/org/folio/linked/data/service/validation/validators/LccnValidator.java
```diff
@@ -0,0 +1,56 @@
...
+import org.folio.linked.data.domain.dto.ResourceRequestDto;
+import org.folio.linked.data.exception.ValidationException;
+import org.folio.linked.data.service.validation.LdValidator;
+import org.folio.linked.data.service.validation.ValidationContext;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Component;
+
+@Component
+@Qualifier(LccnValidator.CODE)
+public class LccnValidator implements LdValidator<ResourceRequestDto> {
```
AndreiBordak, line 16: name looks a bit generic, we have other validations to be added soon, e.g. lccn uniqueness, so i'd suggest to rename it for example to LccnStructureValidator

src/main/java/org/folio/linked/data/configuration/ValidatorsConfiguration.java
AndreiBordak: Regarding the scenario that PK mentioned i think it's a question to PO. If settings are unavailable, should we tell a user that we prevent resource creation/update or we can disable validations for some time? How other modules handle such case?
From technical perspective both cases are fine for us, we can use @Retryable, @Recover annotations and add some logic around them to implements both scenarios.
@pkjacob

askhat-abishev:
> Hi @PBobylev, @askhat-abishev, @AndreiBordak, In order to make the DTO validation approach consistent, I think it will be good to refactor the current `PrimaryTitleDtoValidator` to follow the approach in this PR. What do you think?

Hi @pkjacob 
I agree but, in my opinion, it should be done in the scope of another ticket

src/main/java/org/folio/linked/data/configuration/ValidatorsConfiguration.java
askhat-abishev: @pkjacob we can implement any behavior I believe but first we have to agree on the requirements that should be reflected in the ticket in scope of which we are going to implement the configuration part

src/main/java/org/folio/linked/data/service/validation/LdValidator.java
```diff
@@ -0,0 +1,29 @@
...
+
+  default void check(ValidationContext<T> context) {
+    validate(context);
+    checkErrors(context);
+  }
+
+  private void checkErrors(ValidationContext<T> context) {
+    var validationErrors = context.getValidationErrors();
+    if (!CollectionUtils.isEmpty(validationErrors)) {
+      throw new ValidationException("message", "key", "value");
```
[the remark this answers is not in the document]
askhat-abishev: @pkjacob this code is currently unreachable
see comment on line 28 of ```LccnPatternValidator```

src/main/java/org/folio/linked/data/configuration/ValidatorsConfiguration.java
askhat-abishev: done

src/main/java/org/folio/linked/data/service/validation/validators/LccnValidator.java
askhat-abishev: done

src/main/java/org/folio/linked/data/configuration/ValidatorsConfiguration.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.folio.linked.data.domain.dto.ResourceRequestDto;
+import org.folio.linked.data.service.validation.LdValidator;
+import org.folio.linked.data.service.validation.validators.LccnPatternValidator;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class ValidatorsConfiguration {
```
PBobylev, line 13: **Guys, why are we going to take a step back from the most modern approach to validation based on [Jakarta Bean Validation 3.0](https://beanvalidation.org/)?**

Already existed examples are: 
org.folio.linked.data.validation.PrimaryTitleConstraint
org.folio.linked.data.validation.entity.PrimaryTitleEntityValidator
org.folio.linked.data.validation.dto.PrimaryTitleDtoValidator

**Why are we going to manually build a chain of validators, manually divide them into groups for create/update, etc?**

All of that and even more is possible to implement with Jakarta Bean Validation, it's very agile and it brings less manual work and lower coupling.

Few examples with our needs:
1) Configurable enabled/disabled validations: We could make a validator a Spring bean and inject some validation configuration provider into it
2) Group validations by create/update/etc operations: [Validator groups](https://medium.com/javarevisited/javax-validation-and-groups-acd9a1fb1091) does it for us
3) Gathering a list of validation violations: provided out of the box

PBobylev (changes requested):

src/main/java/org/folio/linked/data/configuration/ValidatorsConfiguration.java
PBobylev: Configuration is out of scope of this ticket, right?
For future: It should be request-scoped I believe, otherwise we can't guarantee that settings are current.

PR #45, opened 2024-11-18
Modld 589

PBobylev:


src/main/java/org/folio/linked/data/controller/advice/ApiExceptionHandler.java
```diff
@@ -39,136 +25,80 @@
...
-      .errors(List.of(new Error()
-        .message(e.getMessage())
-        .type(e.getClass().getSimpleName())
-        .code(code.getValue())))
-      .totalRecords(1);
-    return buildResponseEntity(errorResponse, status);
+    return genericServerExceptionMapper.errorResponseEntity(exception);
   }
 
   private static ResponseEntity<ErrorResponse> buildResponseEntity(ErrorResponse errorResponse, HttpStatus status) {
```
pkjacob, line 104: This method is not used any more.

PBobylev: Fixed, thank you!

src/main/resources/ValidationMessages.properties
```diff
@@ -0,0 +1,2 @@
+primaryTitleConstraint.message=required_primary_main_title
+lccnPatternConstraint.message=lccn_does_not_match_pattern
```
pkjacob, line 2: Minor -> we can change the code to `invalid_lccn_format`

src/main/java/org/folio/linked/data/controller/advice/ApiExceptionHandler.java
```diff
@@ -39,140 +24,80 @@
...
-    logException(exception);
-    return buildResponseEntity(exception, BAD_REQUEST, VALIDATION_ERROR);
+    MissingServletRequestParameterException e) {
+    logException(e);
+    return genericBadRequestMapper.errorResponseEntity(e);
+  }
+
+  @ExceptionHandler(RequestProcessingException.class)
+  public ResponseEntity<ErrorResponse> handleRequestProcessingException(RequestProcessingException exception) {
+    return requestProcessingExceptionMapper.errorResponseEntity(exception);
```
pkjacob, line 94: Call `logException(exception);` here also.

pkjacob (approved):

src/main/java/org/folio/linked/data/controller/advice/ApiExceptionHandler.java
PBobylev: I think we shouldn't do it:
RequestProcessingException is our exception, we throw it manually and we log it together with throwing.
While other exceptions are automatic, that's why we log all of them here.

askhat-abishev (approved):

PR #48, opened 2024-11-20
MODLD-582: LCCN Validation | Ensure uniquely assigned

AndreiBordak:


src/main/java/org/folio/linked/data/client/SearchClient.java
```diff
@@ -0,0 +1,14 @@
+package org.folio.linked.data.client;
+
+import org.folio.linked.data.domain.dto.SearchResponseTotalOnly;
+import org.springframework.cloud.openfeign.FeignClient;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+
+@FeignClient(name = "search")
```
AndreiBordak, line 9: @pkjacob we need to double check the name

src/main/java/org/folio/linked/data/service/search/InstanceSearchServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+      return new ResponseEntity<>(new SearchResponseTotalOnly().totalRecords(0L), HttpStatus.NOT_FOUND);
+    }
+  }
+
+  private String getLccnQuery(Collection<String> lccnCol) {
+    var orLccnQuery = lccnCol.stream()
+      .filter(Objects::nonNull)
+      .map(lccn -> format(LCCN_EQUALS, lccn))
+      .collect(joining(" or "));
+    return format("(%s) and (%s)", orLccnQuery, excludeSuppressedRecords());
```
AndreiBordak, line 53: the logic was copied from https://github.com/folio-org/ui-quick-marc/blob/master/src/hooks/useValidation/validators.js#L445

pkjacob:
Hi @AndreiBordak,
Let us make following 2 changes in the Module descriptor.
1. Add `search.instances.collection.get` permission to the modulePermissions of POST /resource and PUT /resource/id API. (As GET /search/instance API is protected by this permission. See [here](https://github.com/folio-org/mod-search/blob/master/descriptors/ModuleDescriptor-template.json#L127) 
2. Add following to the "requires" array of module descriptor
    {
      "id": "search",
      "version": "1.3"
    }

pkjacob:
Hi @AndreiBordak,
There are some sonar violations. Please take a look

src/main/java/org/folio/linked/data/service/search/InstanceSearchServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+
+  @Override
+  public ResponseEntity<SearchResponseTotalOnly> searchByLccn(Collection<String> lccn) {
+    return search(getLccnQuery(lccn));
+  }
+
+  @Override
+  public ResponseEntity<SearchResponseTotalOnly> searchByLccnExcludingId(Collection<String> lccn, String instanceId) {
+    var lccnQuery = getLccnQuery(lccn);
+    if (isNull(instanceId)) {
```
pkjacob, line 34: This `if` check can be done before computing the value of `lccnQuery` right?

src/main/java/org/folio/linked/data/service/search/InstanceSearchServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+      return searchByLccn(lccn);
+    }
+    return search(format("%s and %s", lccnQuery, format(ID_NOT_EQUALS, instanceId)));
+  }
+
+  private ResponseEntity<SearchResponseTotalOnly> search(String query) {
+    try {
+      return searchClient.searchInstances(query);
+    } catch (FeignException.NotFound e) {
+      return new ResponseEntity<>(new SearchResponseTotalOnly().totalRecords(0L), HttpStatus.NOT_FOUND);
```
pkjacob, line 44: Let us ask Doug if we should return an error to user in this case.

src/main/java/org/folio/linked/data/service/search/InstanceSearchServiceImpl.java
AndreiBordak: fixed.

src/main/java/org/folio/linked/data/service/search/InstanceSearchService.java
```diff
@@ -0,0 +1,12 @@
+package org.folio.linked.data.service.search;
+
+import java.util.Collection;
+import org.folio.linked.data.domain.dto.SearchResponseTotalOnly;
+import org.springframework.http.ResponseEntity;
+
+public interface InstanceSearchService {
+
+  ResponseEntity<SearchResponseTotalOnly> searchByLccn(Collection<String> lccn);
```
PBobylev, line 9: Service layer should not be aware about transport-specific things like Request/Response objects, it's violation of SRP?
Best practice is:
- Happy path:
1) Service contains only "service" logic and returns a DTO.
2) Controller wraps it in a Response object.
- Unhappy path:
1) Service throws an exception (RequestProcessingException in our case)
2) ControllerAdvice wraps it in a Error Response object.
---
Most important benefit, but not the only one: service logic is not messed with transport logic and could be re-used for different sources (Kafka for example)

PBobylev (changes requested):

src/main/java/org/folio/linked/data/service/search/InstanceSearchService.java
AndreiBordak: Fixed.

src/main/java/org/folio/linked/data/controller/advice/ArgumentsInitBinder.java
```diff
@@ -0,0 +1,32 @@
...
+import jakarta.servlet.http.HttpServletRequest;
+import org.folio.linked.data.domain.dto.ResourceRequestDto;
+import org.springframework.web.bind.WebDataBinder;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.InitBinder;
+
+@ControllerAdvice
+public class ArgumentsInitBinder {
+
+  @InitBinder
```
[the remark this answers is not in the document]
AndreiBordak: "value" not working with @RequestBody, moved init binder to controller.

src/main/java/org/folio/linked/data/validation/dto/LccnValidator.java
```diff
@@ -0,0 +1,17 @@
...
+
+import static org.apache.commons.collections4.CollectionUtils.isEmpty;
+
+import jakarta.validation.ConstraintValidator;
+import java.lang.annotation.Annotation;
+import org.folio.linked.data.domain.dto.LccnRequest;
+
+public abstract class LccnValidator<A extends Annotation, T> implements ConstraintValidator<A, T> {
+
+  protected boolean isCurrent(LccnRequest lccnRequest) {
```
PBobylev, line 11: [Let's not use inheritance if it's possible to use composition.](https://en.wikipedia.org/wiki/Composition_over_inheritance)

Here we could just extract a method as a static one to Validation util or smth

PBobylev:
There are [4 New issues](https://sonarcloud.io/project/issues?id=org.folio%3Amod-linked-data&pullRequest=48&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true) according to SQ report above. Let's fix'em at once please?

src/main/java/org/folio/linked/data/service/search/InstanceSearchServiceImpl.java
```diff
@@ -0,0 +1,52 @@
...
+      return searchByLccn(lccn);
+    }
+    return search(format("%s and %s", getLccnQuery(lccn), format(ID_NOT_EQUALS, instanceId)));
+  }
+
+  private SearchResponseTotalOnly search(String query) {
+    try {
+      return searchClient.searchInstances(query).getBody();
+    } catch (Exception e) {
+      throw new RuntimeException("Unable to reach search service");
```
PBobylev, line 41: Throwing generic exception is a bad practice. Why don't we throw failedDependencyException right here?

src/main/java/org/folio/linked/data/validation/dto/LccnUniquenessValidator.java
```diff
@@ -0,0 +1,83 @@
...
+    return ofNullable(findInstanceWithLccn(lccn, inventoryId))
+      .map(SearchResponseTotalOnly::getTotalRecords)
+      .map(count -> count == 0)
+      .orElse(false);
+  }
+
+  private SearchResponseTotalOnly findInstanceWithLccn(List<String> lccn, String inventoryId) {
+    try {
+      return instanceSearchService.searchByLccnExcludingId(lccn, inventoryId);
+    } catch (Exception e) {
```
PBobylev, line 58: we already catch and wrap feign exceptions inside of instanceSearchService, so I don't think we need to catch anything here. Do you see a reason?

src/main/java/org/folio/linked/data/service/search/InstanceSearchServiceImpl.java
AndreiBordak: Because search service is dependency for validation and might be a dependency for other components. If we throw failedDependencyException right here we couldn't know for which component dependency is failed, e.g. 
LccnValidation ->  InstancesSearch (search throws exception and we can say to user that validation failed because of search, otherwise we could only say 'Search is unavailable' without reference to validation which might be confusing)

src/main/java/org/folio/linked/data/validation/dto/LccnUniquenessValidator.java
AndreiBordak: it's needed to build clear message for user that validation failed due to search service.

src/main/java/org/folio/linked/data/validation/dto/LccnValidator.java
AndreiBordak: moved to LccnUtils

AndreiBordak:
> 4 New issues

fixed all except:
image

src/main/java/org/folio/linked/data/service/search/InstanceSearchServiceImpl.java
AndreiBordak: Replaced RuntimeException with SearchException

AndreiBordak: ended up removing the exception throwing.

src/main/java/org/folio/linked/data/controller/advice/ApiExceptionHandler.java
```diff
@@ -88,6 +89,18 @@ public ResponseEntity<ErrorResponse> handleMissingServletRequestParameterExcepti
     return genericBadRequestMapper.errorResponseEntity(e);
   }
 
+  @ExceptionHandler(ValidationException.class)
+  public ResponseEntity<ErrorResponse> handleValidationException(ValidationException exception) {
+    return handleValidationExceptionCause(exception);
+  }
+
+  private ResponseEntity<ErrorResponse> handleValidationExceptionCause(ValidationException exception) {
```
pkjacob, line 97: We do not need this private method right?Better to include this code in the public handleValidationException method itself?

src/main/java/org/folio/linked/data/validation/dto/LccnValidator.java
```diff
@@ -0,0 +1,8 @@
+package org.folio.linked.data.validation.dto;
+
+import jakarta.validation.ConstraintValidator;
+import java.lang.annotation.Annotation;
+
+public abstract class LccnValidator<A extends Annotation, T> implements ConstraintValidator<A, T> {
```
pkjacob, line 6: Any need to keep this interface? May be remove it if it is not needed any more?

src/main/java/org/folio/linked/data/validation/dto/LccnUniquenessValidator.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+
+  private boolean hasCurrentLccn(InstanceField instance) {
+    return getLccnRequest(instance).anyMatch(LccnUtils::isCurrent);
+  }
+
+  private List<String> getLccnValues(InstanceField instance) {
+    return getLccnRequest(instance)
+      .map(LccnRequest::getValue)
```
pkjacob, line 72: Only the current LCCNs has to be checked for duplicates. So, we can do a filter here

```
    return getLccnRequest(instance)
      .filter(LccnUtils::isCurrent)
      .map(....
```

AndreiBordak: Done.

src/main/java/org/folio/linked/data/validation/dto/LccnValidator.java
AndreiBordak: done.

PBobylev:
> > 4 New issues
> 
> fixed all except: image

Let's ignore it via SuppressWarnings too, it's definitely a false-positive

pkjacob (approved):

src/test/java/org/folio/linked/data/e2e/resource/ResourceControllerITBase.java
```diff
@@ -445,6 +552,8 @@ void update_shouldReturn400_ifLccnIsInvalid() throws Exception {
       .content(
         OBJECT_MAPPER.writeValueAsString(updateDto).replaceAll(WORK_ID_PLACEHOLDER, work.getId().toString())
       );
+    when(searchClient.searchInstances(any()))
```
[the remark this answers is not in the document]
AndreiBordak: not sure if it specified, please double check, probably commented by mistake

src/main/java/org/folio/linked/data/controller/advice/ApiExceptionHandler.java
```diff
@@ -88,6 +89,18 @@ public ResponseEntity<ErrorResponse> handleMissingServletRequestParameterExcepti
     return genericBadRequestMapper.errorResponseEntity(e);
   }
 
+  @ExceptionHandler(ValidationException.class)
```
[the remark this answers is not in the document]
AndreiBordak: Yes, it's for case when our validators throw exception

src/test/java/org/folio/linked/data/validation/dto/LccnUniquenessValidatorTest.java
```diff
@@ -0,0 +1,175 @@
...
+    // given
+    var requestDto = new ResourceRequestDto()
+      .id(123L)
+      .resource(new InstanceField().instance(createInstanceRequest(createLccnRequest())));
+    when(folioMetadataRepository.findInventoryIdById(123L)).thenReturn(Optional.empty());
+    when(instanceSearchService.searchByLccnExcludingId(any(), any()))
+      .thenReturn(new SearchResponseTotalOnly().totalRecords(1L));
+
+    // when
+    boolean isValid = validator.isValid(requestDto, null);
```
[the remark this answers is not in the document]
AndreiBordak: wanted to keep explicit boolean here.

src/test/java/org/folio/linked/data/validation/dto/LccnUniquenessValidatorTest.java
```diff
@@ -0,0 +1,175 @@
...
+    var requestDto = new ResourceRequestDto()
+      .id(123L)
+      .resource(new InstanceField().instance(createInstanceRequest(createLccnRequest())));
+    when(folioMetadataRepository.findInventoryIdById(123L))
+      .thenReturn(Optional.of(() -> "6dcb9a08-9884-4a15-b990-89c879a8e999"));
+    when(instanceSearchService.searchByLccnExcludingId(any(), any()))
+      .thenReturn(new SearchResponseTotalOnly().totalRecords(1L));
+
+    // when
+    boolean isValid = validator.isValid(requestDto, null);
```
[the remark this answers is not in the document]
AndreiBordak: same as above.

src/test/java/org/folio/linked/data/validation/dto/LccnUniquenessValidatorTest.java
```diff
@@ -0,0 +1,175 @@
...
+    // given
+    var requestDto = new ResourceRequestDto()
+      .id(123L)
+      .resource(new InstanceField().instance(createInstanceRequest(createLccnRequest())));
+    when(folioMetadataRepository.findInventoryIdById(123L)).thenReturn(Optional.empty());
+    when(instanceSearchService.searchByLccnExcludingId(any(), any()))
+      .thenReturn(new SearchResponseTotalOnly().totalRecords(0L));
+
+    // when
+    boolean isValid = validator.isValid(requestDto, null);
```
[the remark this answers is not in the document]
AndreiBordak: same as above.

src/main/java/org/folio/linked/data/exception/SearchException.java
```diff
@@ -0,0 +1,8 @@
+package org.folio.linked.data.exception;
+
+public class SearchException extends RuntimeException {
```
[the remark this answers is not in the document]
AndreiBordak: good catch, thank you

PBobylev (approved):

PR #52, opened 2024-11-21
MODLD-602 & FOLIO-3678: Use GitHub Workflows api-lint and api-schema-lint and api-doc

pkjacob:
Making similar changes as in https://github.com/folio-org/mod-search/pull/363

### DIFF
--- a/.github/workflows/api-doc.yml
+++ b/.github/workflows/api-doc.yml
@@ -18,9 +18,8 @@ name: api-doc
 
 env:
   API_TYPES: 'OAS'
-  API_DIRECTORIES: 'target/api/openapi'
-  API_INPUT_SPEC: 'src/main/resources/swagger.api/mod-linked-data.yaml'
-  API_EXCLUDES: 'src/main/resources/swagger.api/folio-modules.yaml'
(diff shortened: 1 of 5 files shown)


AndreiBordak (approved):

dcrossleyau:
The "API_EXCLUDES" is incorrect for both api-lint.yml and api-doc.yml -- it is "file name", so remove the path component. That exclusion is not needed in api-schema-lint.yml because it processes JSON files only.

I do not know anything about these modified Workflows that mod-search has made to the defaults linked from https://dev.folio.org/guides/api-doc/#github-workflows (i.e. those at https://github.com/folio-org/.github/tree/master/workflow-templates). So if there are issues then you will need assistance from them. Note that these mod-linked-data workflow files are not those that you say that you utilised from my original mod-search PR 363. Also i see that the "on ... paths" triggers have been removed.

Also remember that the main issue described at MODLD-602 (Invalid JSON pointer) will also arise here.

You should activate the workflows on this branch. With my original PR i did that via an inconsequential edit to an API description file.

dcrossleyau (approved):
I see that the api-lint Workflow run was okay.

The api-doc Workflow will run when you merge to mainline.

PR #64, opened 2024-12-02
MODLD-586: LCCN Normalization | Auto-add white spaces

askhat-abishev:


src/main/java/org/folio/linked/data/validation/dto/LccnPatternValidator.java
```diff
@@ -42,6 +47,16 @@ private boolean isLccnFormatValidationEnabled() {
       .orElse(false);
   }
 
+  private void normalize(LccnRequest lccnRequest) {
+    var normalizer = lccnNormalizers.stream()
+      .reduce(Function.identity(), Function::andThen);
```
PBobylev, line 52: Fix new Sonar violation here please: https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=64&id=org.folio%3Amod-linked-data&open=AZOHXhzDIzf0MBDHogwN

src/test/java/org/folio/linked/data/preprocessing/lccn/SpaceAdderStructureaTest.java
```diff
@@ -0,0 +1,34 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.regex.Pattern;
+import org.folio.spring.testing.type.UnitTest;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+@UnitTest
+class SpaceAdderStructureaTest {
```
PBobylev, line 12: Structurea isn't that readable. Let's name it like StructureA here and in StructureB test too?

src/main/java/org/folio/linked/data/validation/dto/LccnPatternValidator.java
askhat-abishev: fixed

src/test/java/org/folio/linked/data/preprocessing/lccn/SpaceAdderStructureaTest.java
askhat-abishev: I have named these classes like so because checkstyle doesn't allow to call them like SpaceAdderStructure**A**Test/SpaceAdderStructure**B**Test

src/main/java/org/folio/linked/data/preprocessing/lccn/LccnNormalizer.java
```diff
@@ -0,0 +1,15 @@
+package org.folio.linked.data.preprocessing.lccn;
+
+@FunctionalInterface
+public interface LccnNormalizer<T> {
```
pkjacob, line 4: Do we need to overcomplicate this by adding `<T>`? LCCN is always going to be a string. So, instead of `T`, we can hardcode the type as `String` right? Or do you foresee any need to support other datatypes in future?

src/main/java/org/folio/linked/data/validation/dto/LccnPatternValidator.java
```diff
@@ -21,10 +21,18 @@ public class LccnPatternValidator implements ConstraintValidator<LccnPatternCons
   private static final Pattern LCCN_STRUCTURE_B_PATTERN = Pattern.compile("( {2}|[a-z][ a-z])\\d{10}");
 
   private final SpecProvider specProvider;
+  private final LccnNormalizer<String> lccnNormalizer;
+
+  public LccnPatternValidator(SpecProvider specProvider, List<LccnNormalizer<String>> lccnNormalizers) {
+    this.specProvider = specProvider;
+    this.lccnNormalizer = lccnNormalizers.stream().reduce(LccnNormalizer.identity(), LccnNormalizer::andThen);
```
pkjacob, line 28: Hi @askhat-abishev,
Eventhough this code will work, I got little confused in the beginning. This gave me an impression that first we will apply StructureA normalizer and then pass the output to StructureB normalizer. This is not correct. We have to apply either one of the normalizer based on the pattern of the incoming LCCN.

So, I think a more clean structure for LccnNormalizer is as follows

```
public interface LccnNormalizer extends Predicate<String>, UnaryOperator<String> {
  // Return normalized LCCN value if it is valid, otherwise return empty Optional
  default Optional<String> normalize(String t) {
    if (this.test(t)) {
      return Optional.of(this.apply(t));
    }
    return Optional.empty();
  }
}
```

Then apply the normalization in `LccnPatternValidator` as follows
```
  private String normalize(String lccn) {
    return lccnNormalizers
      .stream()
      .flatMap(normalizer -> normalizer.normalize(lccn).stream())
      .findFirst()
      .orElse(lccn);
  }
```

src/main/java/org/folio/linked/data/preprocessing/lccn/impl/AbstractSpaceAdder.java
```diff
@@ -0,0 +1,41 @@
+package org.folio.linked.data.preprocessing.lccn.impl;
+
+import static com.github.jknack.handlebars.internal.lang3.StringUtils.SPACE;
+
+import java.util.regex.Pattern;
+import org.folio.linked.data.preprocessing.lccn.LccnNormalizer;
+
+public abstract class AbstractSpaceAdder implements LccnNormalizer<String> {
```
pkjacob, line 8: minor - `AbstractLccnNormalizer` is a better name I think.
Similarly, `LccnNormalizerStructrueA` and `LccnNormalizerStructrueB`

src/test/java/org/folio/linked/data/preprocessing/lccn/SpaceAdderStructureaTest.java
PBobylev: Checkstyle makes it worse, that's a pity.

src/main/java/org/folio/linked/data/preprocessing/lccn/LccnNormalizer.java
askhat-abishev: Yes, I thought that in future we can normalize something other than just String. It should have been called just ```Normalizer``` but I forgot to rename it properly.

src/main/java/org/folio/linked/data/preprocessing/lccn/impl/AbstractSpaceAdder.java
askhat-abishev: done

pkjacob (approved):

PR #66, opened 2024-12-04
MODLD-594: Resource date fields

AndreiBordak:
https://github.com/folio-org/lib-linked-data-dictionary/pull/21


src/main/resources/changelog/scripts/v-1.0.0/resource_graph/tables/resources.sql
```diff
@@ -10,7 +10,12 @@ CREATE TABLE if not exists resources (
   updated_event_id bigint null,
   active_event_id bigint null,
   index_date timestamp,
-  active boolean default true not null
+  active boolean default true not null,
```
pkjacob, line 13: Moving forward, we have to use ALTER TABLE scripts; Else, deployment to sprint testing env will fail. Earlier, we were in control of the environments, so we could drop the database when such changes are made in DB tables.

AndreiBordak: moved to separate migration.

src/main/java/org/folio/linked/data/model/entity/Resource.java
```diff
@@ -193,6 +219,9 @@ void prePersist() {
       throw new IllegalStateException("Cannot save resource [" + id + "] with types " + types + ". "
         + "Folio metadata can be set only for instance and authority resources");
     }
+    if (isNull(this.createdDate)) {
```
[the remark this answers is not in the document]
AndreiBordak: it will not pass nullable = false constraint if we remove this code.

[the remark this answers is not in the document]
AndreiBordak: Possible solution is @Generated(event = INSERT) annotation, but not in this case, we can't use it due to our graph update process. When we replace a resource by breaking edges and deleting the old one, we want to keep creation date from old resource.

pkjacob:
Hi @AndreiBordak,
We do not have to programmatically populate these `created_*` and `updated_**` fields. Spring provides annotations and classes like @EnableJpaAuditing, AuditingEntityListener, @CreatedBy, @CreatedDate etc which does this automatically. Let us use that feature.

Please refer the following classes in mod-consortia

[AuditableEntity](https://github.com/folio-org/mod-consortia/blob/master/src/main/java/org/folio/consortia/domain/entity/base/AuditableEntity.java)
[FolioAuditorAware](https://github.com/folio-org/mod-consortia/blob/master/src/main/java/org/folio/consortia/config/FolioAuditorAware.java)

You can see similar implementation in some other modules also. See here -> https://github.com/search?q=org%3Afolio-org+%40EnableJpaAuditing&type=code

src/main/java/org/folio/linked/data/service/resource/impl/ResourceServiceImpl.java
```diff
@@ -101,6 +101,8 @@ private Resource getResource(Long id) {
   private Resource saveNewResource(ResourceRequestDto resourceDto, Resource old) {
     var mapped = resourceDtoMapper.toEntity(resourceDto);
     metadataService.ensure(mapped, old.getFolioMetadata());
+    mapped.setCreatedDate(old.getCreatedDate());
+    mapped.setVersion(old.getVersion() + 1);
```
PBobylev, line 105: What about setting createdBy/updatedBy?

PBobylev, line 105: And what about Create operation, don't we want to set createdBy there?

PBobylev (changes requested):

AndreiBordak:
> Hi @AndreiBordak, We do not have to programmatically populate these `created_*` and `updated_**` fields. Spring provides annotations and classes like `@EnableJpaAuditing`, `AuditingEntityListener`, `@CreatedBy`, `@CreatedDate` etc which does this automatically. Let us use that feature.
> 
> Please refer the following classes in mod-consortia [AuditableEntity](https://github.com/folio-org/mod-consortia/blob/master/src/main/java/org/folio/consortia/domain/entity/base/AuditableEntity.java), [FolioAuditorAware](https://github.com/folio-org/mod-consortia/blob/master/src/main/java/org/folio/consortia/config/FolioAuditorAware.java)
> 
> See this commit -> [folio-org/mod-consortia@ed1692a#diff-03919f66fdf8a3c852c986be0af89be443011690729b681baf60e1949849c0ee](https://github.com/folio-org/mod-consortia/commit/ed1692a2d3ca5e938a0efc6b164676aeff7d5a91#diff-03919f66fdf8a3c852c986be0af89be443011690729b681baf60e1949849c0ee)
> 
> You can see similar implementation in some other modules also. See here -> https://github.com/search?q=org%3Afolio-org+%40EnableJpaAuditing&type=code

Hi @pkjacob , for updateDate we're using ```@UpdateTimestamp```, but i'm not sure we can apply @CreateTimestamp because of how we update a resource in ResourceServiceImpl#updateResource.
https://github.com/folio-org/mod-linked-data/pull/66#discussion_r1871781672

src/main/java/org/folio/linked/data/integration/kafka/listener/handler/SourceRecordDomainEventHandler.java
```diff
@@ -78,11 +78,21 @@ private void saveAuthorities(SourceRecordDomainEvent event) {
...
       log.info(EVENT_SAVED, event.getId(), MARC_AUTHORITY, id);
     }
   }
 
+  private void setUserMetadata(Resource resource, SourceRecordDomainEvent event) {
+    if (CREATED == event.getEventType()) {
+      resource.setCreatedBy(event.getEventMetadata().getCreatedBy());
+    }
+    if (UPDATED == event.getEventType()) {
+      resource.setUpdatedBy(event.getEventMetadata().getCreatedBy());
```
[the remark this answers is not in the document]
AndreiBordak: no, createdBy() returns id of the user who created CREATE or UPDATE event

src/main/java/org/folio/linked/data/service/resource/impl/ResourceServiceImpl.java
AndreiBordak: thought it was out of the scope of the story, let me check how we can do this using AuditorAware.

src/main/java/org/folio/linked/data/integration/kafka/listener/handler/SourceRecordDomainEventHandler.java
```diff
@@ -78,11 +78,21 @@ private void saveAuthorities(SourceRecordDomainEvent event) {
...
     if (CREATED == event.getEventType() || UPDATED == event.getEventType()) {
+      setUserMetadata(resource, event);
       var id = resourceMarcService.saveMarcResource(resource);
       log.info(EVENT_SAVED, event.getId(), MARC_AUTHORITY, id);
     }
   }
 
+  private void setUserMetadata(Resource resource, SourceRecordDomainEvent event) {
+    if (CREATED == event.getEventType()) {
+      resource.setCreatedBy(event.getEventMetadata().getCreatedBy());
```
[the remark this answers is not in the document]
AndreiBordak: fixed.

src/main/java/org/folio/linked/data/service/resource/impl/ResourceServiceImpl.java
AndreiBordak: it's out of the scope of the story, isn't it?

src/main/java/org/folio/linked/data/configuration/json/deserialization/event/SourceRecordDomainEventDeserializer.java
```diff
@@ -44,9 +46,24 @@ public SourceRecordDomainEvent deserialize(JsonParser jp, DeserializationContext
     if (node.has(EVENT_PAYLOAD)) {
       event.setEventPayload(getSourceRecord(node.get(EVENT_PAYLOAD)));
     }
+    if (node.has(EVENT_METADATA)) {
```
pkjacob, line 49: We should not use the date coming in the event. Instead, we should use the current date and user.

src/main/java/org/folio/linked/data/integration/kafka/listener/handler/SourceRecordDomainEventHandler.java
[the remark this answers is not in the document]
PBobylev: Do you mean in case of UPDATE event getUpdatedBy is empty and getCreatedBy is used anyway?

src/main/java/org/folio/linked/data/service/resource/impl/ResourceServiceImpl.java
PBobylev: I see it in the table in the ticket. But there is also statement: "The scope of this card is to add date fields to the resource table". So let's clarify with Doug, right? Otherwise there is a controversy and we can't say if we implemented the task or not.

src/main/java/org/folio/linked/data/model/entity/Resource.java
```diff
@@ -193,6 +226,9 @@ void prePersist() {
...
+      this.createdDate = new Timestamp(System.currentTimeMillis());
```
askhat-abishev, line 230: now we don't need this code do we?
aren't we setting the ```createdDate``` using ```touchForCreate``` method?

src/main/java/org/folio/linked/data/configuration/audit/LinkedDataAuditorAware.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.Optional;
+import java.util.UUID;
+import lombok.RequiredArgsConstructor;
+import org.folio.spring.FolioExecutionContext;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.data.domain.AuditorAware;
+import org.springframework.data.jpa.repository.config.EnableJpaAuditing;
+
+@Configuration
+@EnableJpaAuditing(modifyOnCreate = false)
```
askhat-abishev, line 12: why ```false```?

AndreiBordak: because we don't want updatedDate to be modified when a resource is updating via breaking edged and replacing the old resource

src/main/java/org/folio/linked/data/model/entity/Resource.java
AndreiBordak: not in all cases, we need it for the case with resource update

src/main/java/org/folio/linked/data/configuration/audit/LinkedDataAuditEntityListener.java
```diff
@@ -0,0 +1,28 @@
...
+import lombok.AllArgsConstructor;
+import org.folio.linked.data.model.entity.Resource;
+import org.springframework.beans.factory.ObjectFactory;
+import org.springframework.data.auditing.AuditingHandler;
+import org.springframework.data.jpa.domain.support.AuditingEntityListener;
+import org.springframework.stereotype.Component;
+
+@Component
+@AllArgsConstructor
+public class LinkedDataAuditEntityListener extends AuditingEntityListener {
```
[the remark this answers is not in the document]
AndreiBordak: yes, we need that for the case with resource update, the code is modified a bit by introducing the following check:
```if (target instanceof Resource resource && isNull(resource.getCreatedBy())) { ...```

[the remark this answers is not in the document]
Comment (AndreiBordak) [reply src/main/java/org/folio/linked/data/configuration/audit/LinkedDataAuditEntityListener.java:16]: > yes, we need that for the case with resource update, the code is modified a bit by introducing the following check: `if (target instanceof Resource resource && isNull(resource.getCreatedBy())) { ...`
it's need for the case with a resource update

File: src/main/java/org/folio/linked/data/configuration/audit/LinkedDataAuditEntityListener.java [hunk 10]
```diff
@@ -0,0 +1,28 @@
...
+
+  private ObjectFactory<AuditingHandler> handler;
+
+  @Override
+  @PrePersist
+  public void touchForCreate(Object target) {
+    if (target instanceof Resource resource && isNull(resource.getCreatedBy())) {
+      ofNullable(handler.getObject())
```
pkjacob, line 24: `handler` is nullable. So, `handler.getObject()` may through NullPointerException. The `ofNullable(handler.getObject())` will not help.

May be it be better to do like this?
```
    if (target instanceof Resource resource && isNull(resource.getCreatedBy())) {
      super.touchForCreate(target);
    }
```

src/main/java/org/folio/linked/data/service/resource/ResourceServiceImpl.java
```diff
@@ -103,6 +105,10 @@ private Resource saveNewResource(ResourceRequestDto resourceDto, Resource old) {
     var mapped = resourceDtoMapper.toEntity(resourceDto);
     metadataService.ensure(mapped, old.getFolioMetadata());
     resourceEdgeService.copyOutgoingEdges(old, mapped);
+    mapped.setCreatedDate(old.getCreatedDate());
+    mapped.setVersion(old.getVersion() + 1);
+    mapped.setCreatedBy(old.getCreatedBy());
+    mapped.setUpdatedBy(folioExecutionContext.getUserId());
```
pkjacob, line 111: Shouldn't we set updatedDate here with the current date & time?
Consider this scenario:
Day 1: A new resource created using POST API or import
Day 2: The resource is updated through PUT API. The updatedDate should contain the current date.

AndreiBordak: updated_date will contain the current date, ```@UpdateTimestamp``` will do this for us.

src/main/java/org/folio/linked/data/configuration/audit/LinkedDataAuditEntityListener.java
AndreiBordak: have updated the code

src/main/java/org/folio/linked/data/configuration/audit/LinkedDataAuditorAware.java
pkjacob: We want updatedDate to be modified in this case right?
Consider this scenario:
Day 1: A resource got created.
Day 2: The resource got updated using PUT api. We want the update date to be modified with the current date.

src/main/resources/changelog/scripts/v-1.0.0/resource_graph/migration/resources_date_fields.sql
```diff
@@ -0,0 +1,12 @@
+alter table resources
+  add column created_date timestamp default current_timestamp not null,
+  add column updated_date timestamp default current_timestamp not null,
```
pkjacob, line 3: is `updated_date` not null? I think updated_date will be null for new resources. updated_date will be populated when the resource is first updated. Can we do that?

src/main/resources/changelog/scripts/v-1.0.0/resource_graph/migration/resources_date_fields.sql
```diff
@@ -0,0 +1,12 @@
+alter table resources
+  add column created_date timestamp default current_timestamp not null,
```
pkjacob, line 2: As created_date is a "not null" field, should we add a update script to set current_timestamp for all existing records? Else, upgrade will fail for sprint testing env right?

Another option is to ask Oleksandr Haimanov if he can delete mod-linked-data from sprint testing environment so that we can start from scratch. Let me send him a message including you.

pkjacob: See Oleksandr's response in slack. This is not a problem. We have to open a ticket to delete mod-linked-data from the environments.

Also, we do not have  to create alter_table SQL. Instead, we can include it in the Create table script as you originally did.

src/main/resources/changelog/scripts/v-1.0.0/resource_graph/migration/resources_date_fields.sql
AndreiBordak: as per the story requirements when a cataloger saves a Work resource for the first time, the created_date and updated_date fields are auto-populated with the same value

src/main/java/org/folio/linked/data/configuration/audit/LinkedDataAuditorAware.java
AndreiBordak: sorry for confusion, meant updatedBy of course, updated_date will be updated anyway.
if we have modifyOnCreate = true then updatedBy will be overwrited  with value from createdBy on graph update after we manually set createdBy/updatedBy, we don't want to do that in this case.

src/main/resources/changelog/scripts/v-1.0.0/resource_graph/migration/resources_date_fields.sql
AndreiBordak: have reverted the changes

askhat-abishev (approved):

src/test/java/org/folio/linked/data/e2e/resource/ResourceControllerITBase.java
```diff
@@ -124,16 +124,20 @@
 import static org.folio.linked.data.test.TestUtil.OBJECT_MAPPER;
 import static org.folio.linked.data.test.TestUtil.SIMPLE_WORK_WITH_INSTANCE_REF_SAMPLE;
 import static org.folio.linked.data.test.TestUtil.WORK_WITH_INSTANCE_REF_SAMPLE;
+import static org.folio.linked.data.test.TestUtil.assertResourceMetadata;
```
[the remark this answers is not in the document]
AndreiBordak: https://github.com/folio-org/mod-linked-data/pull/78

PBobylev (approved):

PR #69, opened 2024-12-06
Save AdminMetadata edge out of incoming SourceRecordDomainEvent

PBobylev:
https://folio-org.atlassian.net/browse/MODLD-592


src/main/java/org/folio/linked/data/integration/kafka/listener/handler/SourceRecordDomainEventHandler.java
```diff
@@ -41,11 +45,8 @@ public void handle(SourceRecordDomainEvent event, SourceRecordType recordType) {
     }
     if (recordType == MARC_AUTHORITY) {
       saveAuthorities(event);
-      //  } else if (recordType == MARC_BIB) {
-      //     For now, you can discard the event if it is a MARC_BIB record
-      //     In coming sprints, we will have stories to update the hrid (extracted out of MARC field 001) in DB for marc
-      //     bib records
-      //     this.updateHrId(event);
+    } else if (recordType == MARC_BIB) {
```
pkjacob, line 48: I think it will be safer to add one more filtering here for eventType == UPDATED as we do not have to consider newly created BIB records for admin data updates. What you think?

ie, 

```
if (recordType == MARC_BIB && event.getEventType() == UPDATED) {
   ...
}
```

PBobylev: Actually it's CREATE operation, isn't it?
First time when we save a record in LDE and it's sent to Inventory/SRS, it will be sent back with CREATE operation type.
But I thought what if HRID (001) is updated, when we better handle UPDATE operations too. But I've checked it and there is no possibility for updating HRID in the Inventory, so let's limit it with CREATE, right?

pkjacob (approved):

src/main/java/org/folio/linked/data/integration/kafka/listener/SourceRecordDomainEventListener.java
```diff
@@ -57,7 +58,7 @@ private void processRecord(ConsumerRecord<String, SourceRecordDomainEvent> consu
       .map(SourceRecordType::fromValue)
       .orElseThrow();
 
-    if (sourceRecordType == MARC_AUTHORITY) {
+    if (sourceRecordType == MARC_AUTHORITY || sourceRecordType == MARC_BIB) {
```
askhat-abishev, line 61: we can completely remove this check aren't we?

askhat-abishev (approved):

src/main/java/org/folio/linked/data/integration/kafka/listener/SourceRecordDomainEventListener.java
PBobylev: nope, there are more types in SRS, holdings for example

PR #71, opened 2024-12-09
MODLD-613: Delete kafka topics after deleting tenant

pkjacob:


src/main/java/org/folio/linked/data/service/tenant/LinkedTenantService.java
askhat-abishev, line 1: looks like service should be renamed to ```LinkedDataTenantService```

askhat-abishev (approved):

PBobylev (approved):

PR #93, opened 2025-01-15
MODLD-642: Retain edges when resource is fetched from DB

pkjacob:


src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
```diff
@@ -83,7 +82,6 @@ private Resource createResourceFromSrs(String srsId) {
         .flatMap(this::contentAsJsonString)
         .flatMap(this::firstAuthorityToEntity)
         .map(resourceGraphService::saveMergingGraph)
-        .map(Resource::copyWithNoEdges)
```
pkjacob, line 86: Hi @AndreiBordak, Do you remember why `.map(Resource::copyWithNoEdges)` was added here? Do you see any reason for that?

src/test/java/org/folio/linked/data/e2e/resource/ResourceControllerUpdateWorkIT.java
```diff
@@ -0,0 +1,207 @@
...
+
+  private Resource getInstance(Resource work) {
+    var title = new Resource()
+      .addTypes(ResourceTypeDictionary.TITLE)
+      .setDoc(getDoc("""
+        {
+          "http://bibfra.me/vocab/marc/mainTitle": ["simple_instance"]
+        }
+        """))
+      .setLabel("simple_work");
```
askhat-abishev, line 130: simple_instance

askhat-abishev (approved):

src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
AndreiBordak: This logic was inherites from somewhere PK, tbh i don't remember.

AndreiBordak (approved):

PR #102, opened 2025-01-27
MODLD-650: Optimize log messages

pkjacob:


askhat-abishev (approved):

src/main/java/org/folio/linked/data/integration/kafka/sender/inventory/InstanceCreateMessageSender.java
```diff
@@ -21,6 +22,7 @@
 import org.springframework.context.annotation.Profile;
 import org.springframework.stereotype.Service;
 
+@Log4j2
```
PBobylev, line 25: There are 4 classes there we use Slf4j instead of Log4j2:
- CoreMapperImpl
- CacheCleaningJob
- ResourceServiceImpl
- SpecProviderImpl

Might be we could also fix it here?

PBobylev (approved):

PR #104, opened 2025-01-28
MODLD-646: Update the response structure of authorityAssignmentCheck API

pkjacob:
Should be merged along with https://github.com/folio-org/ui-linked-data/pull/84

### DIFF
--- a/src/main/resources/swagger.api/mod-linked-data.yaml
+++ b/src/main/resources/swagger.api/mod-linked-data.yaml
@@ -263,12 +263,11 @@ paths:
               $ref: schema/assignmentCheckDto.json
       responses:
         '200':
-          description: Returns plain text boolean as response
+          description: Returns the result of the assignment check, including whether the assignment is valid and, if not, the reason for invalidity
           content:
-            text/plain:
+            application/json:
               schema:
-                type: string
-                example: true|false
+                $ref: schema/assignmentCheckResponseDto.json
         '400':
           $ref: '#/components/responses/badRequestResponse'
         '500':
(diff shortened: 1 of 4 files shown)


src/main/resources/swagger.api/mod-linked-data.yaml
```diff
@@ -267,8 +267,7 @@ paths:
           content:
             text/plain:
```
PBobylev, line 268: Now it should be application/json, right?
Description is outdated too (line 266)

src/main/resources/swagger.api/schema/assignmentCheckResponseDto.json
```diff
@@ -0,0 +1,19 @@
...
+  "type": "object",
+  "properties": {
+    "validAssignment": {
+      "type": "boolean",
+      "description": "Indicates if the assignment is valid"
+    },
+    "reason": {
+      "type": "string",
+      "enum": ["UNSUPPORTED_MARC", "NO_LCCN", "NOT_VALID_FOR_TARGET"],
+      "description": "Indicates the reason why the assignment is not valid. Only present if 'validAssignment' is false."
```
PBobylev, line 13: Such description makes me feel that parameter name isn't so descriptive.
What if we name it like invalidationReason?

PBobylev (changes requested):

src/main/resources/swagger.api/schema/assignmentCheckResponseDto.json
pkjacob: Good point. Copilot suggested the parameter name 'invalidAssignmentReason'. So, I used it :)

src/main/resources/swagger.api/mod-linked-data.yaml
pkjacob: Yes. Fixed.

PBobylev (approved):

PR #106, opened 2025-01-30
MODLD-650: Include tenant ID, request ID, user ID & module ID in log messages; Optimize logs for event consumption.

pkjacob:
**Current Logs:** (No tenant ID, user ID etc in log message)
Screenshot 2025-01-30 at 1 05 18 PM

**After the change:**

Normal Logs:
Screenshot 2025-01-30 at 12 54 45 PM

JSON logs:
Screenshot 2025-01-30 at 12 49 47 PM

### DIFF
--- a/src/main/resources/log4j2-json.properties
+++ b/src/main/resources/log4j2-json.properties
@@ -1,3 +1,6 @@
+name = PropertiesConfig
+packages = org.folio.spring.logging
+
 appenders = console
 
 appender.console.type = Console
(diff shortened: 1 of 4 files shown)


src/main/resources/log4j2.properties
```diff
@@ -1,32 +1,23 @@
 appenders = console
```
pkjacob, line 1: Copied from https://github.com/folio-org/folio-spring-support/blob/709129462a8b4f094454d734ef8a91dd55483387/folio-spring-base/src/main/resources/log4j2.properties

src/main/resources/log4j2-json.properties
```diff
@@ -1,3 +1,6 @@
+name = PropertiesConfig
```
pkjacob, line 1: Copied from https://github.com/folio-org/folio-spring-support/blob/709129462a8b4f094454d734ef8a91dd55483387/folio-spring-base/src/main/resources/log4j2-json.properties

askhat-abishev (approved):

PBobylev (approved):

PR #107, opened 2025-01-31
MODLD-644: Preservation of MARC | MARC derived record

askhat-abishev:


src/main/java/org/folio/linked/data/mapper/kafka/inventory/InstanceIngressMessageMapper.java
```diff
@@ -40,7 +41,7 @@ public abstract class InstanceIngressMessageMapper {
 
   protected String toMarcJson(Resource resource) {
     var resourceModel = resourceModelMapper.toModel(resource);
-    return ld2MarcMapper.toMarcJson(resourceModel);
+    return ld2MarcMapper.toMarcJson(resourceModel, UnmappedMarcHandling.MERGE);
```
PBobylev, line 44: we don't have to pass this argument and keep using the old method, right?

askhat-abishev: right
removed

pkjacob (approved):

PR #113, opened 2025-02-10
MODLD-654: Retain Resource properties that are not rendered in the UI

askhat-abishev:


src/main/java/org/folio/linked/data/service/resource/ResourceServiceImpl.java
```diff
@@ -39,6 +55,21 @@
 @RequiredArgsConstructor
 public class ResourceServiceImpl implements ResourceService {
 
+  private static final Set<String> INSTANCE_PROPERTIES = Set.of(
```
pkjacob, line 58: Better to move this to a separate class.
May be a new ResourceProperties service?

src/main/java/org/folio/linked/data/service/resource/ResourceServiceImpl.java
```diff
@@ -116,6 +148,7 @@ private Resource saveNewResource(ResourceRequestDto resourceDto, Resource old) {
     resourceEdgeService.copyOutgoingEdges(old, mapped);
```
pkjacob, line 148: We use a separate service `resourceEdgeService` for copying outgoing edges. However, we copy unmapped marc and properties in this same service.

Maybe create a separate service for this purpose ?
```
interface ResouceCopyService {
   public void copyEdgesAndProperties(Resource old, Resource new); // copy edges, properties & unmapped marc in this method.
}
```


Then we will invoke `resouceCopyService.copyEdgesAndProperties(old, mapped);` from here.

src/main/java/org/folio/linked/data/service/resource/ResourceServiceImpl.java
pkjacob, line 58: Instead of two separate sets, we could create a Map like we do in ResourceEdgeServiceImpl

`private static final Map<ResourceTypeDictionary, Set<String>> PROPERTIES_TO_BE_COPIED = ....;`

src/main/java/org/folio/linked/data/service/resource/ResourceServiceImpl.java
askhat-abishev: Yes, ideally it should be a separate service but since it's a temporary code a did it that way.

src/main/java/org/folio/linked/data/service/resource/ResourceServiceImpl.java
```diff
@@ -146,4 +179,29 @@ private void copyUnmappedMarc(Resource old, Resource mapped) {
...
+      .stream()
+      .filter(entry -> {
+        if (from.isOfType(INSTANCE)) {
+          return INSTANCE_PROPERTIES.contains(entry.getKey());
+        } else if (from.isOfType(ResourceTypeDictionary.WORK)) {
+          return WORK_PROPERTIES.contains(entry.getKey());
+        } else {
+          return false;
+        }
+      })
```
AndreiBordak, line 202: filtering condition might be extracted to a separate method for better readability

src/main/java/org/folio/linked/data/service/resource/ResourceServiceImpl.java
```diff
@@ -146,4 +179,29 @@ private void copyUnmappedMarc(Resource old, Resource mapped) {
...
   }
 
+  @SneakyThrows
+  private void copyProperties(Resource from, Resource to) {
+    if (from.getDoc() == null) {
+      return;
+    }
+
+    var fromDoc = objectMapper.treeToValue(from.getDoc(), new TypeReference<HashMap<String, List<String>>>() {});
+    var toDoc = to.getDoc() == null
```
AndreiBordak, line 189: toDoc initialization looks to early. I'd suggest to find satisfying entries first and then put them all to a new map:
```
var entries = fromDoc.entrySet()
      .stream()
      .filter(entry -> {
        if (from.isOfType(INSTANCE)) {
          return INSTANCE_PROPERTIES.contains(entry.getKey());
        } else if (from.isOfType(ResourceTypeDictionary.WORK)) {
          return WORK_PROPERTIES.contains(entry.getKey());
        } else {
          return false;
        }
      })
      .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));

      // check entries exists then init new map
      toDoc.putAll(entries);
```

src/main/java/org/folio/linked/data/service/resource/ResourceServiceImpl.java
askhat-abishev: done

src/main/java/org/folio/linked/data/service/resource/ResourceServiceImpl.java
askhat-abishev: done

src/main/java/org/folio/linked/data/service/resource/ResourceServiceImpl.java
askhat-abishev: done

src/main/java/org/folio/linked/data/service/resource/copy/ResourceCopyServiceImpl.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  private List<Map.Entry<String, List<String>>> getProperties(Resource from) throws JsonProcessingException {
+    var fromDoc = objectMapper.treeToValue(from.getDoc(), new TypeReference<HashMap<String, List<String>>>() {});
+    var fromType = from.getTypes()
+      .stream()
+      .findFirst()
+      .orElseThrow()
+      .getUri();
+    return fromDoc.entrySet()
```
pkjacob, line 96: minor - I think it will be little more efficient (& readable) if we do like this
```
var propertiesToCopy = PROPERTIES_TO_BE_COPIED.getOrDefault(fromType, emptySet());
return fromDoc.entrySet()
      .stream()
      .filter(entry -> propertiesToCopy.contains(entry.getKey()))
      .toList();
```

pkjacob (approved):

src/main/java/org/folio/linked/data/service/resource/copy/ResourceCopyServiceImpl.java
askhat-abishev: done

PR #119, opened 2025-02-20
MODLD-617: special handling when authority is reverted to a previous …

pkjacob:
…state

### DIFF
--- a/src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
+++ b/src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
@@ -126,18 +126,44 @@ public class ResourceMarcAuthorityServiceImpl implements ResourceMarcAuthoritySe
   @Override
   public Long saveMarcAuthority(org.folio.ld.dictionary.model.Resource modelResource) {
     var mapped = resourceModelMapper.toEntity(modelResource);
-    if (!mapped.isAuthority()) {
-      var message = "Resource is not an authority";
-      log.error(message);
-      throw new IllegalArgumentException(message);
-    }
-    if (resourceRepo.existsById(modelResource.getId())) {
+    validateResource(mapped);
+    if (sameResourceExistsByIdAndSrsId(mapped)) {
       return updateAuthority(mapped);
     }
-    if (folioMetadataRepository.existsBySrsId(modelResource.getFolioMetadata().getSrsId())) {
+    if (resourceExistsBySrsId(mapped)) {
       return replaceAuthority(mapped);
     }
     return createAuthority(mapped);
+  }
+
+  private static void validateResource(Resource resource) {
+    if (!resource.isAuthority()) {
+      logAndThrow("Resource is not an authority");
+    }
+
+    if (resource.getFolioMetadata() == null || resource.getFolioMetadata().getSrsId() == null) {
+      logAndThrow("SRS ID is missing in the resource");
+    }
+  }
+
+  private static void logAndThrow(String message) {
+    log.error(message);
+    throw new IllegalArgumentException(message);
+  }
+
+  private boolean sameResourceExistsByIdAndSrsId(Resource resource) {
+    var id = resource.getId();
+    var srsId  = resource.getFolioMetadata().getSrsId();
+    return resourceRepo.existsById(id) && getIdBySrsId(srsId).filter(id::equals).isPresent();
+  }
+
+  private Optional<Long> getIdBySrsId(String srsId) {
+    return folioMetadataRepository.findIdBySrsId(srsId)
+      .map(FolioMetadataRepository.IdOnly::getId);
+  }
+
+  private boolean resourceExistsBySrsId(Resource resource) {
+    return folioMetadataRepository.existsBySrsId(resource.getFolioMetadata().getSrsId());
   }
 
   private Long updateAuthority(Resource resource) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
```diff
@@ -176,9 +202,6 @@ private void logMarcAction(Resource resource, String existence, String action) {
   private Long saveAndPublishEvent(Resource resource, Function<Resource, ResourceEvent> resourceEventSupplier) {
     var newResource = resourceGraphService.saveMergingGraph(resource);
     var event = resourceEventSupplier.apply(newResource);
-    if (event instanceof ResourceReplacedEvent rre) {
```
pkjacob, line 179: `rre.previous()` is already a managed entity (as it was just retrieved from the DB). So, no need to explicitly save

src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
```diff
@@ -126,20 +126,46 @@ private RequestProcessingException notFoundException(String srsId) {
   @Override
   public Long saveMarcAuthority(org.folio.ld.dictionary.model.Resource modelResource) {
     var mapped = resourceModelMapper.toEntity(modelResource);
-    if (!mapped.isAuthority()) {
-      var message = "Resource is not an authority";
-      log.error(message);
-      throw new IllegalArgumentException(message);
-    }
-    if (resourceRepo.existsById(modelResource.getId())) {
```
pkjacob, line 134: Scenario:
1. Via MARC authority, create a authority record  with just 100$a - Example: `100 $a Jane Doe`
2. Go back to authority and add a $d to the Person authority - Example: `100 $a Jane Doe $d 1900`
3. Again update the authority and remove $d so that authority is reverted to the original state. - Example: `100 $a Jane Doe`

Bug:
After step 2, two resources exists in the DB
1. ID:1 -- John Doe (isPreferred=false, srsId=null)
2. ID:2 -- John Doe, 1900  (isPreferred=true, srsId=abcd)

At step 3, saveMarcAuthority method receive John Doe  with ID 1 as input.
Earlier, we first did `if (resourceRepo.existsById(modelResource.getId()))  { return updateAuthority(..) }`
the isPreferred flag of ID 1 got changed to true. However, resource with ID 2 was untouched. So, it also retained the sPreferred=true flag and srsId.

src/test/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImplTest.java
```diff
@@ -152,8 +155,8 @@ void fetchAuthorityOrCreateFromSrsRecord_shouldThrowNotFound_ifRecordNotExistsIn
     when(exceptionBuilder.notFoundSourceRecordException(any(), any())).thenReturn(emptyRequestProcessingException());
 
     // then
-    assertThrows(RequestProcessingException.class,
-      () -> resourceMarcAuthorityService.fetchAuthorityOrCreateFromSrsRecord(agent));
+    assertThatThrownBy(() -> resourceMarcAuthorityService.fetchAuthorityOrCreateFromSrsRecord(agent))
```
pkjacob, line 158: switching to assertJ to be consistent with other assertions in this file.

src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
```diff
@@ -125,19 +125,48 @@ private RequestProcessingException notFoundException(String srsId) {
...
+
+  private static void logAndThrow(String message) {
+    log.error(message);
+    throw new IllegalArgumentException(message);
+  }
+
+  private boolean existsBySrsIdAndIdUnchanged(Resource resource) {
+    var id = resource.getId();
+    var srsId  = resource.getFolioMetadata().getSrsId();
+    return resourceRepo.existsById(id) && getIdBySrsId(srsId).filter(id::equals).isPresent();
```
askhat-abishev, line 160: I'm a bit confused with the method name because here we check ```resourceRepo.existsById(id)```

askhat-abishev (approved):

src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
PBobylev, line 160: What if instead of calling resourceRepo.existsById and then getIdBySrsId(srsId) to check equality, we just write a new method FolioMetadataRepository#existsByIdAndSrsId and call it?

src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
PBobylev: What if an Authority exists without srsId? I am not sure, is that possible at all?
If yes we will never update it with incoming authority, right?

pkjacob: SRS ID should be always present. 
I added code to throw IllegalArgumentException just in case if SRS ID is not present. See line 150

src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
pkjacob: Hi, Actually, the check `resourceRepo.existsById(id)` was unnecessary. So, I refactored the method.

pkjacob: You are right.
I refactored the method. But with a slightly different way.

1. Get the Optional<resource ID of the existing resource> from DB using the given SRS ID.
2. If resource ID is empty, then create the new resource in graph.
3. Else, check if the existing resource ID is same as the id of the incoming resource. If yes, update the existing resource.
4. Else, create a new resource and make the old one obsolete.

PBobylev (approved):

PR #159, opened 2025-04-08
MODLD-703: Reduce unnecessary log messages

askhat-abishev:


src/main/java/org/folio/linked/data/job/CacheCleaningJob.java
```diff
@@ -15,12 +17,18 @@ public class CacheCleaningJob {
...
   @CacheEvict(value = SETTINGS_ENTRIES, allEntries = true)
   @Scheduled(fixedRateString = "${mod-linked-data.cache.ttl.settings-entries}")
   public void emptySettingsEntries() {
-    log.info("Emptying {} cache", SETTINGS_ENTRIES);
+    log.info(EMPTY_CACHE_MSG, SETTINGS_ENTRIES);
+  }
+
+  @CacheEvict(value = MODULE_STATE, allEntries = true)
+  @Scheduled(fixedRateString = "${mod-linked-data.cache.ttl.module-state}")
+  public void emptyModuleState() {
```
pkjacob, line 31: Hi @askhat-abishev, 
Can we invoke this method from the `afterTenantUpdate` and `afterTenantDeletion` methods of `LinkedDataTenantService`?

This will ensure that the cache is cleared after a tenant is created or deleted.

src/main/java/org/folio/linked/data/integration/kafka/listener/SourceRecordDomainEventListener.java
```diff
@@ -43,7 +46,15 @@ public class SourceRecordDomainEventListener {
     concurrency = "#{folioKafkaProperties.listener['source-record-domain-event'].concurrency}",
     topicPattern = "#{folioKafkaProperties.listener['source-record-domain-event'].topicPattern}")
   public void handleSourceRecordDomainEvent(List<ConsumerRecord<String, SourceRecordDomainEvent>> consumerRecords) {
-    consumerRecords.forEach(this::processRecord);
+    consumerRecords.forEach(consumerRecord -> {
+      var tenant = getHeaderValueByName(consumerRecord, TENANT).orElseThrow();
```
pkjacob, line 50: minor -> Is there a good exception that we can throw here if the tenant ID is missing in kafka header?
This is minor as the chances of this happening is very low in production.

src/main/java/org/folio/linked/data/job/CacheCleaningJob.java
askhat-abishev: @pkjacob do you mean to remove the scheduled cleanup and instead perform the cleanup in ```LinkedDataTenantService``` or add cleanup to ```LinkedDataTenantService``` in addition to scheduled one?

src/main/java/org/folio/linked/data/integration/kafka/listener/SourceRecordDomainEventListener.java
askhat-abishev: @pkjacob in my opinion, ```IllegalArgumentException``` can be used.

src/main/java/org/folio/linked/data/job/CacheCleaningJob.java
askhat-abishev: done

AndreiBordak (approved):

src/main/java/org/folio/linked/data/service/LinkedDataApplicationService.java
```diff
@@ -0,0 +1,32 @@
...
+    FROM pg_tables
+    WHERE tablename = 'resources'
+    AND schemaname = '%TENANT%_mod_linked_data'
+    ) AS table_existence""";
+
+  private final JdbcTemplate jdbcTemplate;
+
+  @Override
+  @Cacheable(cacheNames = MODULE_STATE)
+  public boolean isModuleInstalled(String tenant) {
```
PBobylev, line 26: There is already existed functionality in folio-spring-base:
org.folio.spring.service.TenantService#tenantExists, so we could re-use org.folio.linked.data.service.tenant.LinkedDataTenantService and drop this class, right?

PBobylev (changes requested):

src/main/java/org/folio/linked/data/service/LinkedDataApplicationService.java
askhat-abishev: done

pkjacob (approved):

PR #162, opened 2025-04-14
MODLD-707: Optimize log messages

askhat-abishev:


src/main/java/org/folio/linked/data/util/ResourceUtils.java
```diff
@@ -152,4 +153,12 @@ public static boolean isPreferred(Resource resource) {
       .filter(value -> value.equals("true"))
       .isPresent();
   }
+
+  public static String getType(Resource resource) {
+    return resource.getTypes()
+      .stream()
+      .map(ResourceTypeEntity::getUri)
+      .findFirst()
```
pkjacob, line 161: Let us return all the types from here. Method name also can be changed as `public static List<String> getTypes(...)`

askhat-abishev: done

AndreiBordak (approved):

PBobylev (approved):

pkjacob (approved):

PR #165, opened 2025-04-16
MODLD-682: Solve loop that appears while updating a Person to a Person, Concept

AndreiBordak:


src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
```diff
@@ -171,19 +172,30 @@ private Long updateAuthority(Resource resource) {
...
-        logMarcAction(resource, "not found by id, but found by srsId [" + srsId + "]",
-          "be saved as a new version of previously existed resource [id " + previous.getId() + "]");
-        return saveAndPublishEvent(resource, saved -> new ResourceReplacedEvent(previousObsolete, saved.getId()));
-      })
+      .map(previous -> markObsoleteAndReplace(previous, resource))
       .orElseThrow(() -> notFoundException(srsId));
   }
 
+  private Long markObsoleteAndReplace(Resource previous, Resource incoming) {
+    if (notEqual(previous.getTypes(), incoming.getTypes())) {
```
pkjacob, line 180: Hi @AndreiBordak,

Some of these steps should still execute even when the types differ.

Something like -
```
if (equal(revious.getTypes(), incoming.getTypes())) {
  // do everything that was done previously
} else {
    markObsolete(previous); // previous resource should be marked obsolete as the underlying authority record has updated.
    setPreferred(incoming, true); // we should make the new resource preferred
    save(incoming); // save. But do not publish any events. `ResourceReplacedEvent` is anyway ignored for authority records
   // also add an appropriate log
}
```

AndreiBordak: @pkjacob Should we save both `previous` and `incoming`?

src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
```diff
@@ -171,19 +172,34 @@ private Long updateAuthority(Resource resource) {
...
+      var previousObsolete = markObsolete(previous);
+      setPreferred(incoming, true);
+      addReplacedByRelation(previousObsolete, incoming);
+      logMarcAction(incoming,
+        "not found by id, but found by srsId [" + incoming.getFolioMetadata().getSrsId() + "]",
+        "be saved as a new version of previously existed resource [id " + previous.getId() + "]");
+      return saveAndPublishEvent(incoming, saved -> new ResourceReplacedEvent(previousObsolete, saved.getId()));
+    }
+    markObsolete(previous);
+    resourceGraphService.saveMergingGraph(previous);
```
pkjacob, line 190: `previous` is already a managed object right?
So, what is the point in calling `      addReplacedByRelation(previousObsolete, incoming);`?

src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
```diff
@@ -171,19 +172,34 @@ private Long updateAuthority(Resource resource) {
...
+    setPreferred(incoming, true);
+    var saved = resourceGraphService.saveMergingGraph(incoming);
+    logMarcAction(incoming, "set as preferred", "be saved");
```
pkjacob, line 193: We should log that the old resource is marked obsolete.

src/test/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImplTest.java
```diff
@@ -203,12 +204,38 @@ void saveMarcAuthority_shouldUpdateAuthority_ifGivenModelExistsByIdAndSrsId() {
     verify(applicationEventPublisher).publishEvent(new ResourceUpdatedEvent(mapped));
   }
 
+  @Test
```
pkjacob, line 207: Will this be enough to test the end to end flow?
Or should we add an IT in ResourceMarcAuthorityServiceIT?

src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
pkjacob: previous is already a managed Entity right? So, why should we save it again?

src/main/java/org/folio/linked/data/service/resource/marc/ResourceMarcAuthorityServiceImpl.java
```diff
@@ -171,27 +173,42 @@ private Long updateAuthority(Resource resource) {
...
+    if (Objects.equals(previous.getTypes(), incoming.getTypes())) {
+      var previousObsolete = markObsolete(previous);
+      setPreferred(incoming, true);
+      addReplacedByRelation(previousObsolete, incoming);
+      logMarcAction(incoming,
+        "not found by id, but found by srsId [" + incoming.getFolioMetadata().getSrsId() + "]",
+        "be saved as a new version of previously existed resource [id " + previous.getId() + "]");
+      return saveAndPublishEvent(incoming, saved -> new ResourceReplacedEvent(previousObsolete, saved.getId()));
+    }
+    markObsolete(previous);
```
pkjacob, line 190: I think it may be more readable if we move this to an `else` block.

pkjacob (approved):

PBobylev (approved):

askhat-abishev (approved):

PR #198, opened 2025-05-23
MODLD-740: Define server response to file import

zepheiryan:
https://folio-org.atlassian.net/browse/MODLD-740

Update response to JSON containing a list of resource IDs and a CSV activity log. Also add a mechanism to convert some generic media types to rdf4j-recognized media types.

### DIFF
--- a/src/main/java/org/folio/linked/data/service/rdf/RdfImportServiceImpl.java
+++ b/src/main/java/org/folio/linked/data/service/rdf/RdfImportServiceImpl.java
@@ -1,7 +1,6 @@
 package org.folio.linked.data.service.rdf;
 
 import java.io.IOException;
-import java.util.List;
 import java.util.Set;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.log4j.Log4j2;
@@ -42,10 +41,9 @@ public class RdfImportServiceImpl implements RdfImportService {
   }
 
   private ImportFileResponseDto save(Set<org.folio.ld.dictionary.model.Resource> resources) {
-    ImportFileResponseDto response;
-    String reportCsv = "";
-    ImportUtils.ImportReport report = new ImportUtils.ImportReport();
-    List<Long> ids = resources.stream()
+    var reportCsv = "";
+    var report = new ImportUtils.ImportReport();
+    var ids = resources.stream()
       .map(resourceModelMapper::toEntity)
       .filter(r -> {
         boolean exists = resourceRepo.existsById(r.getId());
@@ -75,9 +73,8 @@ public class RdfImportServiceImpl implements RdfImportService {
     try {
       reportCsv = report.toCsv();
     } catch (IOException e) {
-      log.warn("I/O error while generating CSV report, returning empty report: {}", e);
+      log.warn("I/O error while generating CSV report, returning empty report", e);
     }
-    response = new ImportFileResponseDto(ids, reportCsv);
-    return response;
+    return new ImportFileResponseDto(ids, reportCsv);
   }
 }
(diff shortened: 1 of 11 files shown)


src/main/java/org/folio/linked/data/service/rdf/RdfImportServiceImpl.java
```diff
@@ -30,31 +32,52 @@ public class RdfImportServiceImpl implements RdfImportService {
...
     } catch (IOException e) {
       throw exceptionBuilder.badRequestException("Rdf import incoming file reading error", e.getMessage());
     }
   }
 
-  private List<Long> save(Set<org.folio.ld.dictionary.model.Resource> resources) {
-    return resources.stream()
+  private ImportFileResponseDto save(Set<org.folio.ld.dictionary.model.Resource> resources) {
+    ImportFileResponseDto response;
+    String reportCsv = "";
```
PBobylev, line 46: let's use 'var' keyword when possible, it's part of our code convention

PBobylev (changes requested):

src/main/resources/swagger.api/schema/importFileResponseDto.json
```diff
@@ -0,0 +1,23 @@
...
+    "resources": {
+      "type": "array",
+      "description": "A list of resource identifiers successfully imported",
+      "items": {
+        "type": "integer",
+        "format": "int64"
+      }
+    },
+    "log": {
+      "type": "string",
```
PBobylev, line 15: It's for Excel file downloading, right?
Don't we want to declare csv or excel mime type here and provide respective binary data instead of string?
I could be mistaken, but how client's browser define the data as an Excel file otherwise? Some frontend logic will overwrite a type?

src/main/resources/swagger.api/schema/importFileResponseDto.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Responds to a file import with a list of imported resources and an activity log",
+  "type": "object",
+  "properties": {
+    "resources": {
+      "type": "array",
+      "description": "A list of resource identifiers successfully imported",
```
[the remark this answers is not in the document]
zepheiryan: Just to wrap this up, while the resources list is currently duplicated in both the response and the log contents, we expect to change this in the log to a list of URIs, so ultimately the thin thread response won't contain any duplicated lists.

src/main/java/org/folio/linked/data/service/rdf/RdfImportServiceImpl.java
```diff
@@ -30,31 +32,52 @@ public class RdfImportServiceImpl implements RdfImportService {
...
+              r.getId(),
+              r.getLabel(),
+              ImportUtils.Status.FAILURE,
+              "Already exists in graph"));
         }
         return !exists;
       })
       .map(resource -> {
         metadataService.ensure(resource);
         var saved = resourceGraphService.saveMergingGraph(resource);
```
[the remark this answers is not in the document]
zepheiryan: This is an intentional logging side effect to report certain cases back to the user, I don't think there's much to be gained by refactoring here.

src/main/resources/swagger.api/schema/importFileResponseDto.json
pkjacob: >> In order to accomplish this, the UI client will read in everything, and for this part, it will turn the contents into an anchor and then simulate a click to begin the download.

I like the idea. I think this should be enough for thin thread.

PBobylev: I am not against it, since it's just Thin Thread. Just saying that csv/excel mime type for 'log' and multipart/form-data for whole response should also do the trick. Approved anyway.

PR #200, opened 2025-05-28
MODLD-744: API to return profile metadata array for a resource type

pkjacob:
Database Schema Update:
===
**Profile Table:** Added two new non-null fields — name and resourceType.
**Foreign Key:** resourceType is linked to the type_hash column of the resource_type table.
**JPA Entity Update:** Corresponding changes made in the Profile entity class to reflect the new fields.

Liquibase Migration Process Change:
==
Previously, profile records were inserted using Liquibase SQL scripts. This is not possible moving forward due to the foreign key relation with `resource_type` table.  Records are inserted to `resource_type` table by `DictionaryWorker.java`. We can insert profiles to `profiles` table only after `DictionaryWorker` is executed.

Also the previous approach introduced complexity during development, especially when frequent changes to the profile JSON required schema resets (which involved coordination with the Kitfox team).

**New Approach:** Record insertion is now handled by a Java class - ProfileWorker.

To ensure correct execution order, we’ve added the @Order annotation to all worker classes. `ProfileWorker` is now guaranteed to run after `DictionaryWorker`. 

Note:
One drawback of the new approach is that ProfileWorker will update profile records even when there are no changes. However, this is not a major problem as there be only few profiles (max 5 or 6).

New API:
===
1. Introduced a new endpoint: `GET /linked-data/profiles/metadata`
2. Updated the module descriptor to include a new section for `linked-data-profiles`.

Failing Jenkins build
==
Jenkins build should pass after merging this PR - https://github.com/folio-org/mod-linked-data/pull/202


descriptors/ModuleDescriptor-template.json
```diff
@@ -154,6 +144,27 @@
         }
       ]
     },
+    {
+      "id": "linked-data-profiles",
+      "version": "1.0",
+      "handlers": [
+        {
```
pkjacob, line 151: This API will be removed very soon (after Siarhei's refactoring is complete)

src/main/resources/changelog/scripts/v-2.0.0/metadata/tables/upgrade_profiles_table.sql
```diff
@@ -0,0 +1,21 @@
+--liquibase formatted sql
+
+--changeset upgrade_profiles_table_structure dbms:postgresql
+
+-- Delete all profiles
+delete from profiles;
```
pkjacob, line 6: Deleting all profiles. Profiles will be inserted to DB by ProfileWorker.java

src/main/resources/swagger.api/mod-linked-data.yaml
```diff
@@ -211,6 +211,7 @@ paths:
 
   /linked-data/profile:
     get:
+      deprecated: true
```
pkjacob, line 214: We will remove this API once Siarhei's refactoring is over.

src/main/java/org/folio/linked/data/repo/ProfileRepository.java
```diff
@@ -1,7 +1,12 @@
...
+import java.util.List;
 import org.folio.linked.data.model.entity.Profile;
+import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
+import org.springframework.data.repository.query.Param;
 
-public interface ProfileRepository extends CrudRepository<Profile, String> {
+public interface ProfileRepository extends CrudRepository<Profile, Long> {
+  @Query("SELECT p FROM Profile p WHERE p.resourceType.uri = :resourceTypeUri")
+  List<Profile> findByResourceTypeUri(@Param("resourceTypeUri") String resourceTypeUri);
```
PBobylev, line 11: We don't have to write a query, Spring Data method with according naming could do the same thing without @Query and @Param
`List<Profile> findByResourceTypeUri(String resourceTypeUri);`

src/main/resources/changelog/scripts/v-2.0.0/metadata/changelog.xml
```diff
@@ -3,5 +3,5 @@
                    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
-  <include file="tables/change_profiles_id_to_text.sql" relativeToChangelogFile="true"/>
```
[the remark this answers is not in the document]
PBobylev: Are we going to erase the DB at all snapshot and sprint envs in such case?

PBobylev (changes requested):

src/main/resources/changelog/scripts/v-2.0.0/metadata/changelog.xml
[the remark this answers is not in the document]
pkjacob: Yes. I will create a ticket for Kitfox to do it once.  Only sprint testing envs has to be cleaned. Snapshot environments are created from scratch every day.

src/main/java/org/folio/linked/data/repo/ProfileRepository.java
pkjacob: Done

src/test/java/org/folio/linked/data/e2e/ProfileControllerIT.java
```diff
@@ -12,46 +11,25 @@
...
   @Test
-  void getProfile_shouldReturnProfile() throws Exception {
+  void getProfile_returnsProfile() throws Exception {
     //given
-    var profileEntity = new Profile();
-    profileEntity.setId("2");
-    profileEntity.setValue("[{\"key\": \"BIBFRAME 2.0\"}]");
-    profileRepository.save(profileEntity);
-    var requestBuilder = get(PROFILE_URL)
+    var requestBuilder = get("/linked-data/profile")
```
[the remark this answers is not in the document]
pkjacob: Test will be removed when the deprecated API is removed

zepheiryan (approved):

PR #217, opened 2025-06-10
MODLD-754: Insert simplified profile to DB

pkjacob:
Insert Simplified profile that @SKarolFolio  created for UILD-548 into the DB


src/test/java/org/folio/linked/data/e2e/ProfileControllerIT.java
```diff
@@ -59,7 +59,7 @@ void getProfileById_returnsProfile() throws Exception {
       .andExpect(content().contentType(TEXT_PLAIN_VALUE + ";charset=UTF-8"))
       .andReturn().getResponse().getContentAsString();
 
-    assertTrue(profile.contains("\"id\": \"lc:RT:bf2:Monograph:Work\""));
+    assertTrue(profile.contains("\"id\": \"monograph\""));
```
SKarolFolio, line 62: Same here

SKarolFolio (changes requested):

SKarolFolio (approved):

PR #220, opened 2025-06-12
Avoid join RawMarc table

pkjacob:
Currently, the Resource table includes the following mapping:

```
@OneToOne(cascade = ALL, mappedBy = "resource", orphanRemoval = true)
@PrimaryKeyJoinColumn
private RawMarc unmappedMarc;
```

As a result, a left join to the `raw_marcs` table is always performed when a Resource is fetched. However, this join is often unnecessary, as only Instance resources have corresponding records in the raw_marcs table.

This PR removes the `unmappedMarc` field from the Resource class and handles the saving and retrieval of unmapped MARC data explicitly in code.

### DIFF
--- a/src/main/java/org/folio/linked/data/model/entity/Resource.java
+++ b/src/main/java/org/folio/linked/data/model/entity/Resource.java
@@ -102,10 +102,6 @@ public class Resource implements Persistable<Long> {
   @PrimaryKeyJoinColumn
   private FolioMetadata folioMetadata;
 
-  @OneToOne(cascade = ALL, mappedBy = "resource", orphanRemoval = true)
-  @PrimaryKeyJoinColumn
-  private RawMarc unmappedMarc;
-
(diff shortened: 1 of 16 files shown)


src/main/java/org/folio/linked/data/model/entity/Resource.java
```diff
@@ -102,10 +102,6 @@ public class Resource implements Persistable<Long> {
   @PrimaryKeyJoinColumn
   private FolioMetadata folioMetadata;
 
-  @OneToOne(cascade = ALL, mappedBy = "resource", orphanRemoval = true)
```
pkjacob, line 105: Having this here will result in an unnecessary table join of "raw_marc" table  whenever a resource is fetched. We know that unmappedMarc is present only for Instance resources.

PBobylev (approved):

PR #231, opened 2025-06-23
MODLD-762: RDF Export API

PBobylev:


src/main/java/org/folio/linked/data/controller/ResourceController.java
```diff
@@ -74,6 +77,11 @@ public ResponseEntity<ResourceMarcViewDto> getResourceMarcViewById(Long id) {
     return ResponseEntity.ok(resourceMarcService.getResourceMarcView(id));
   }
 
+  @Override
+  public ResponseEntity<RdfResourceDto> exportInstanceToRdf(Long id) {
```
pkjacob, line 81: I do not think we ned to wrap the RDF in RdfResourceDto.
We can directly return the RDF. 
ResponseEntity<JsonNode> should be enough. right?

PR #254, opened 2025-07-15
MODLD-362: Refactor SingleResourceMapperUnit & implementations

pkjacob:
**Problem:**

When the same resource types can be linked using different predicates, we’re currently forced to create a separate MapperUnit class for each predicate. This leads to unnecessary subclass creation.

For example, the following four classes are identical in structure but differ only in the `predicate` value specified in the `@MapperUnit` annotation:

1. [PublicationMapperUnit.java](https://github.com/folio-org/mod-linked-data/blob/master/src/main/java/org/folio/linked/data/mapper/dto/monograph/instance/sub/provision/PublicationMapperUnit.java)
2. [ManufactureMapperUnit.java](https://github.com/folio-org/mod-linked-data/blob/master/src/main/java/org/folio/linked/data/mapper/dto/monograph/instance/sub/provision/ManufactureMapperUnit.java)
3. [DistributionMapperUnit.java](https://github.com/folio-org/mod-linked-data/blob/master/src/main/java/org/folio/linked/data/mapper/dto/monograph/instance/sub/provision/DistributionMapperUnit.java)
4. [ProductionMapperUnit.java](https://github.com/folio-org/mod-linked-data/blob/master/src/main/java/org/folio/linked/data/mapper/dto/monograph/instance/sub/provision/ProductionMapperUnit.java)

This issue becomes even more worse with "language part" mappings ([MODLD-362](https://folio-org.atlassian.net/browse/MODLD-362)), where we would need to create up to 16 nearly identical mapper units, e.g.:

```
@MapperUnit(..., predicate = SUMMARY_LANGUAGE, ...)
class SummaryLanguageMapperUnit extends AbstractLanguageMapperUnit {
   // simple constructor
}


@MapperUnit(..., predicate = SUNG_OR_SPOKEN_TEXT_LANGUAGE, ...)
class SungOrSpokenTextLanguageMapperUnit extends AbstractLanguageMapperUnit {
   // simple constructor
}

etc.
```
**Solution (After Refactoring):**

The `@MapperUnit` annotation now supports an array of predicate values instead of a single one.
The specific predicate is passed at runtime to the `toDto` method via the `ResourceMappingContext` parameter.

This change allows a single mapper unit to handle multiple predicates, eliminating the need for unnecessary subclasses.

See this [commit](https://github.com/folio-org/mod-linked-data/pull/254/commits/22ecbf4d8e13d982653bfe8b9f65dec98a165943#diff-1a87549d731078be75ed199dabff7c1331340b30ff37dd69891f515b70b34578).


zepheiryan (approved):

src/main/java/org/folio/linked/data/mapper/dto/common/SingleResourceMapperUnit.java
```diff
@@ -1,14 +1,17 @@
...
 public interface SingleResourceMapperUnit {
 
-  <P> P toDto(Resource source, P parentDto, Resource parentResource);
+  <P> P toDto(Resource resourceToConvert, P parentDto, ResourceMappingContext context);
 
   Resource toEntity(Object dto, Resource parentEntity);
 
   Set<Class<?>> supportedParents();
 
+  record ResourceMappingContext(Resource source, Predicate predicate) {
```
PBobylev, line 15: The 'source' here is a parent resource, right?
That's really confusing naming as for me. A 'source' in mapping context usually is a mapping source, an object to be mapped, it's 'resourceToConvert' basically.

But I see you mean 'source' from ResourceEdge, right? So to eliminate any possible confusion, let's get rid of 'source' name here at all?

pkjacob: Renamed "source" to "parentResource"

pkjacob:
PR no longer needed. Changes are included in this PR -> https://github.com/folio-org/mod-linked-data/pull/256

PR #257, opened 2025-07-17
UILD-557: Add support for updated extent handling

zepheiryan:
https://folio-org.atlassian.net/browse/UILD-557

In addition to required profile changes for treating `extent` as an object rather than a literal, adjustments were also made to consolidate on `http://bibfra.me/vocab/lite/extent` and remove the transient `http://bibfra.me/vocab/lite/extentV2` property originally used to support both cases.


pkjacob:
Hi @zepheiryan 

Let us remove Line 502 from this file -> https://github.com/folio-org/lib-linked-data-marc4ld/blob/master/src/main/resources/marc4ld.yml#L502. We no longer want to map 300 $a to EXTENT property of Instance.
 
Also, we can remove EXTENT from the PropertyDictionary -> https://github.com/folio-org/lib-linked-data-dictionary/blob/master/src/main/java/org/folio/ld/dictionary/PropertyDictionary.java#L47. 
EXTENT will be part of only PredicateDictionary moving forward.

PBobylev (approved):

PR #261, opened 2025-07-21
MODLD-786: Validate series required title property

zepheiryan:
https://folio-org.atlassian.net/browse/MODLD-786


src/main/java/org/folio/linked/data/validation/dto/PartOfSeriesTitleDtoValidator.java
```diff
@@ -0,0 +1,39 @@
...
+  ConstraintValidator<PartOfSeriesTitleConstraint, List<PartOfSeries>> {
+
+  @Override
+  public boolean isValid(List<PartOfSeries> series, ConstraintValidatorContext context) {
+    if (isNull(series)) {
+      return true;
+    }
+
+    return series.stream()
+      .filter(this::seriesIsNotEmpty)
```
PBobylev, line 25: What about empty series? Why are we skipping them?
What happens if all the series are empty? The validation passes, but there won't be a title., right?

zepheiryan: Good point, I checked, and using the API directly with an empty series does result in an empty node. The UI does prevent an empty series from being submitted at all, but I went ahead and removed the emptiness check so that a missing title fails regardless of the rest of the property values.

PR #272, opened 2025-07-29
MODLD-799: Code refactoring & cleanup - Use "IdentifierRequest.json" and "IdentifierResponse.json" for all types of Identifiers

pkjacob:
At present, we use separate openapi schema files like `LccnRequest.json`, `IsbnRequest.json`, `EanRequest.json`, `OtherIdResponse.json` etc for different kinds of identifiers. All these models have almost the same fields. 

This PR consolidates all those different models into two models -`IdentifierRequest.json` and `IdentifierResponse.json`

Only `Lccn` and `Isbn` are currently supported by UI. Other identifiers like` Ean`, `OtherId` etc are not supported by UI today. So, the schema changes will not break the UI.

Note - I have not touched `LocalId.json` in this PR. We will take a closer look at the mapping of 019 and deal with it in a separate ticket.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -32,6 +32,7 @@
 - Update profiles to support book format handling [MODLD-726](https://folio-org.atlassian.net/browse/MODLD-726)
 - Enhance `GET import-supported` API to check if MARC record is a Serial [MODLD-809](https://folio-org.atlassian.net/browse/MODLD-809)
 - Conversion rules from MARC to Graph - adaptation [MODLD-805](https://folio-org.atlassian.net/browse/MODLD-805)
+- Standardized all Identifier requests/responses to IdentifierRequest.json and IdentifierResponse.json; removed redundant schemas. [MODLD-799](https://folio-org.atlassian.net/browse/MODLD-799)
 
 
 ## 1.0.4 (04-24-2025)


src/main/resources/swagger.api/schema/resource/response/identifier/IdentifierResponse.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
-  "description": "LCCN Info",
+  "description": "Identifier response DTO",
```
PBobylev, line 3: We could also move 'IdentifierRequest.json' to 'common/identifier/ and re-use it here to replace Name and Status fields. Thus also lets us delete 'IdentifierName.json' file

src/main/resources/swagger.api/schema/resource/request/identifier/IdentifierRequest.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
-  "description": "LCCN Info",
+  "description": "Identifier request DTO",
```
PBobylev, line 3: One thing is confusing here: the name of the file sounds like it's a base for all Identifiers, while it's not.
It's to be used for LCCN and ISSN, right?
I think we better rename it to something specific, like 'NameAndStatusIdentifier' or just 'NameAndStatus'?

pkjacob: Used IdentifierRequest for all requests except "LocalId".

We will take care of "LocalId" as a separate ticket. We need to look into "LocalId" (019) in more detail.

src/main/java/org/folio/linked/data/mapper/dto/monograph/instance/sub/identified/EanMapperUnit.java
```diff
@@ -3,61 +3,62 @@
...
-    return map.isEmpty() ? null : coreMapper.toJson(map);
+  @Override
+  protected ResourceTypeDictionary getIdentifierType() {
+    return ID_EAN;
   }
 
+  @Override
+  protected IdentifierResponse getResponseDto(Resource resourceToConvert) {
+    var identifierResponse = super.getResponseDto(resourceToConvert);
+    identifierResponse.setValue(getPropertyValues(resourceToConvert, EAN_VALUE));
```
pkjacob, line 53: Note -> This overridden method can be removed as part of [MODLD-810](https://folio-org.atlassian.net/browse/MODLD-810). Superclass method can be made private.

pkjacob:
> Does this merit a changelog entry?

Yes. I will add now.

PR #293, opened 2025-08-19
MODLD-826: API for Publication Frequency (Serials)

PBobylev:


src/main/resources/swagger.api/schema/resource/response/PublicationFrequencyResponse.json
```diff
@@ -0,0 +1,24 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Publication frequency DTO",
+  "allOf": [
```
pkjacob, line 4: Hi @PBobylev 
Can we add the following line here?

```  
"x-from-string-field": "label",
```

Reason -
Earlier, we were mapping the MARC 310 to a string `PUBLICATION_FREQUENCY` property of the Instance resource([ref](https://github.com/folio-org/lib-linked-data-marc4ld/blob/v1.0.3/src/main/resources/marc4ld.yml#L502-L508)). We removed this logic now. Libraries that installed Sunflower version of FOLIO will have such string PUBLICATION_FREQUENCY properties in the DB. [CoreMapperImpl -> toDtoWithEdges](https://github.com/folio-org/mod-linked-data/blob/master/src/main/java/org/folio/linked/data/mapper/dto/common/CoreMapperImpl.java#L39) will fail for such instances.

Adding `"x-from-string-field": "label",` will avoid this problem. With this change, the string publication frequency will get mapped to "label" property of the PublicationFrequencyResponse.

pkjacob (approved):

zepheiryan (approved):

PR #305, opened 2025-08-26
MODLD-782: Show the actual domain for the graph in the export file

PBobylev:


src/main/java/org/folio/linked/data/integration/client/ConfigurationClient.java
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.cloud.openfeign.FeignClient;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+
+@FeignClient(name = "configurations")
+public interface ConfigurationClient {
+
+  @SuppressWarnings("java:S7180")
+  @Cacheable(cacheNames = SETTINGS_ENTRIES, key = "#moduleName")
+  @GetMapping("${config.client.path:/entries}?query=module=={moduleName}")
```
pkjacob, line 16: Looking at mod-configuration's README file, I think we can pass the `code` also here.
`query=(module=={moduleName} and code=={code})`
It will be more efficient as we do not have to cache all other configs for the module.

src/main/java/org/folio/linked/data/service/rdf/ResourceUrlProvider.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.function.LongFunction;
+import lombok.RequiredArgsConstructor;
+import org.folio.linked.data.integration.ConfigurationService;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class ResourceUrlProvider implements LongFunction<String> {
+
+  private static final String URL_PATTERN = "%s/linked-data-editor/resources/%s/edit";
```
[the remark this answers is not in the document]
PBobylev: Current URL leads to LDE UI which already lets us to open an instance or a work. And UI forces a user to log-in first.
Do you think we better replace it with our back-end endpoint? But it's not available without auth token and nothing there forces a user to log-in

src/main/java/org/folio/linked/data/integration/client/ConfigurationClient.java
PBobylev: makes sense, done

pkjacob (approved):

PR #308, opened 2025-08-28
MODLD-858: Use different Maps for storing entities with and without ID

pkjacob:
After fixing the problem and testing it, I now realize that it is not  a big problem like what I initially thought. 
In all XYZ mapper units, we were setting the IDs of the leaf nodes first and then setting the ID of the parent nodes. For example, [CategoryMapperUnit.java](https://github.com/folio-org/mod-linked-data/blob/master/src/main/java/org/folio/linked/data/mapper/dto/resource/common/category/CategoryMapperUnit.java) first set the ID of CategorySet resource and then set the fingerprint of Category resource. So, the problem did not occur there.

In [PartOfSeriesMapperUnit.java](https://github.com/folio-org/mod-linked-data/blob/master/src/main/java/org/folio/linked/data/mapper/dto/resource/common/work/sub/PartOfSeriesMapperUnit.java#L83) alone, we set the IDs from the top level first. That uncovered this problem.

Please see if this is a risky change. If it is a risky change, we can update `PartOfSeriesMapperUnit` to set the IDs starting from the leaf nodes first. That should "hide" the problem. (But the problem will still exist in ResourceModelMapper)

Solution
---

1. Use `HashMap` when entity have ID (as we do today). `HashMap` uses the `equals` method, which in turn compare the `id` value.
2. Fallback to `IdentityHashMap` when the entity do not have ID. `IdentityHashMap` uses `==` (reference equality) instead of `.equals()`.


pkjacob:
> @pkjacob I would say our ResourceModelMapper is designed to accept a resource with not empty ID. And that should be guaranteed beforehand. All the relations between resources are traced by IDs plus "caching" (CyclicGraphContext) of already mapped resources is also based on ID.
> 
> I would suggest to keep it as is. What we better do as for me:
> 
> 1. ResourceModelMapper - add a validation of incoming resource graph - thrown an exception if any resource doesn't contain an id
> 2. Cover the issue with an IT involving all existed MapperUnits. It could be a test which scans for all MapperUnit beans, passes to each mapperUnit simple resource with label for example and checks if ID is set
> 
> Doing that we unhide a problem:
> 
> 1. ResourceModelMapper will throw an exception once we try to pass a resource w/o ID
> 2. IT for all MapperUnits adds fail-fast trigger showing us a problem even faster, with no necessity for waiting for an exception in Runtime

Ok. Cancelling this PR. We will create a tech debt ticket.
I will create the PR for [MODLD-807](https://folio-org.atlassian.net/browse/MODLD-807) to work around this problem.

PR #315, opened 2025-09-04
MODLD-865: PreferredProfileService updates existed record instead of creating new one with DataIntegrityViolation

PBobylev:


src/main/resources/log4j2-json.properties
PBobylev, line 1: Code cleaning: the same config file is provided by folio-spring-support: https://github.com/folio-org/folio-spring-support/tree/master?tab=readme-ov-file#logging

src/test/java/org/folio/linked/data/e2e/database/DatabaseITBase.java
```diff
@@ -19,8 +19,6 @@ abstract class DatabaseITBase {
             WHERE table_type = 'BASE TABLE' AND table_schema = ?
     """;
 
-  @Value("${spring.application.name}")
-  private String appName;
```
PBobylev, line 23: Code cleaning: nut used variable

src/main/resources/log4j2-json.properties
PBobylev: I suppose the whole file was copied from Folio Spring support lib a long time ago. Then it probalby became outdated and was updated. We could check it with PK (author of the updates) later.
Reverted for now

PR #330, opened 2025-09-23
MODLD-874: Get resource graph by inventory IDs

pkjacob:
My initial approach was to create the following method in ResourceRepository (as described in the ticket)

`Set<Resource> findByFolioMetadataInventoryIds(Set<String> inventoryIds)`

However, there is a performance issue with this approach. Even though all the resources will be fetched in a single query from the DB, the ResourceModelMapper will later fetch each of the outgoing edges separately from the DB, causing a lot of separate queries against the DB.

To solve this problem, I am using the new view that @zepheiryan  created. I think the code is self explanatory. Please check if this approach looks OK.


zepheiryan:
Looks like a good approach to me.

PBobylev:
I think since we are going to use the endpoint for Authorities fetching only, there should be no or just few outgoing edges.
But thinking about endpoint itself, this approach is more optimal and it's looks good.

PR #543, opened 2026-06-03
MODLD-1069 - Extend existing Integration tests to fully cover TestRail test cases

alekGbuz:
[MODLD-1069](https://folio-org.atlassian.net/browse/MODLD-1069) - Extend existing Integration tests to fully cover TestRail test cases


zepheiryan (approved):

src/test/java/org/folio/linked/data/e2e/mappings/PutResourceIT.java
```diff
@@ -0,0 +1,58 @@
...
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.put;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+import org.folio.linked.data.model.entity.Resource;
+import org.junit.jupiter.api.Test;
+import org.springframework.test.web.servlet.ResultActions;
+
+/**
+ * Extends {@link PostResourceIT} with a PUT round-trip test.
```
PBobylev: Let's name it as PutResourceIT then?

src/test/java/org/folio/linked/data/e2e/mappings/PostAndPutResourceIT.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.folio.linked.data.model.entity.Resource;
+import org.junit.jupiter.api.Test;
+import org.springframework.test.web.servlet.ResultActions;
+
+/**
+ * Extends {@link PostResourceIT} with a PUT round-trip test.
+ * Subclasses that need to verify update (PUT) behaviour should extend this class instead of {@link PostResourceIT}.
+ */
+public abstract class PostAndPutResourceIT extends PostResourceIT {
```
PBobylev: extending a test by another one isn't that good practice. Let's extract common part to an abstract parent. And we already have such, let's use it probably?

mappingProfile/bibframe2.0/instance.json
```diff
@@ -0,0 +1,24 @@
+{
```
PBobylev: Looks like this folder and these files are not needed, right?

PBobylev (changes requested):

src/test/java/org/folio/linked/data/e2e/mappings/PostAndPutResourceIT.java
alekGbuz: Hi @PBobylev , I checked previously that adding put logic to abstract PostResourceIT using the same approach as used for post request - here in case if IT tests already implement PostResourceIT ( there are a lot of them already) it is necessary to provide some implementation of put logic for them as well or make test pass by default without any checks assertion.

mappingProfile/bibframe2.0/instance.json
alekGbuz: @PBobylev removed that directory as it is not needed.

src/test/java/org/folio/linked/data/e2e/mappings/PutResourceIT.java
alekGbuz: @PBobylev renamed it to PutResourceIT

PBobylev (approved):

PR #546, opened 2026-06-12
MODLD-1040: Update profile settings model to include multiple settings per profile and a label

zepheiryan:
https://folio-org.atlassian.net/browse/MODLD-1040

Await front end approval before merging to avoid breaking deploy. Module permission updates require changes in downstream repositories.


src/main/resources/changelog/scripts/v-3.0.0/metadata/tables/upgrade_profile_settings_table.sql
zepheiryan, line 1: @PBobylev This branch is potentially a target for backporting to a 2.x patch release, should this be in v-3.0.0 or moved to v-2.0.0?

PBobylev: That's godd as it's now:
- Firstly we need it to be merged, deployed and tested on current U release branch/env;
- After this we could proceed with backporting it to the T;
- Then again it has to be merged, deployed and tested, but now on T release branch/env;
- Only after such path is complete, it's allowed to release it as part of Trillium CSP;

PBobylev:
> Await front end approval before merging to avoid breaking deploy. Module permission updates require changes in downstream repositories.

So we have to wait for the https://github.com/folio-org/ui-linked-data/pull/418, right?

zepheiryan:
> > Await front end approval before merging to avoid breaking deploy. Module permission updates require changes in downstream repositories.
> 
> So we have to wait for the [folio-org/ui-linked-data#418](https://github.com/folio-org/ui-linked-data/pull/418), right?

Yes, that's the one.

PBobylev (approved):

PR #550, opened 2026-06-16
MODLD-1083: Add Authority profiles 

SKarolFolio:
[https://folio-org.atlassian.net/browse/MODLD-1083](https://folio-org.atlassian.net/browse/MODLD-1083)

### DIFF
--- a/src/main/resources/profiles/authority-family.json
+++ b/src/main/resources/profiles/authority-family.json
@@ -1,0 +1,146 @@
+{
+  "id": 8,
+  "name": "Family",
+  "resourceType": "_authority",
+  "value": [
+    [
+      {
+        "type": "block",
+        "displayName": "Family",
+        "bfid": "lde:Profile:Authority",
+        "uriBFLite": "_authority",
+        "children": [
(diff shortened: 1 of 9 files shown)


src/main/resources/profiles/authority-family.json
```diff
@@ -0,0 +1,146 @@
+{
+  "id": 8,
+  "name": "Family",
+  "resourceType": "_authority",
```
alekGbuz: Hi @SKarolFolio running locally uploading profile with _authority as resourceType got such issue: 
`[ERROR: null value in column "resource_type" of relation "profiles" violates not-null constraint`
The same time it works using this one : 
`http://bibfra.me/vocab/lite/Family`
Should it be changed?

PBobylev: http://bibfra.me/vocab/lite/  + authority name (line 3)

SKarolFolio: Done

PBobylev (approved):

alekGbuz (approved):

PR #551, opened 2026-06-17
MODLD-1083: Update authority profiles

SKarolFolio:
Update authority profiles to simplify the structure by removing unnecessary array nesting.
Add Identifiers section to authority profiles.

[https://folio-org.atlassian.net/browse/MODLD-1083](https://folio-org.atlassian.net/browse/MODLD-1083)

### DIFF
--- a/src/main/resources/profiles/authority-family.json
+++ b/src/main/resources/profiles/authority-family.json
@@ -16,7 +16,8 @@
         "Profile:Authority:MiscellaneousInformation",
         "Profile:Authority:AttributionQualifier",
         "Profile:Authority:FullerFormOfName",
-        "Profile:Authority:Affiliation"
+        "Profile:Authority:Affiliation",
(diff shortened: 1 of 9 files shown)


alekGbuz (approved):

src/main/resources/profiles/authority-family.json
```diff
@@ -3,144 +3,142 @@
   "name": "Family",
```
PBobylev, line 3: Please also add a section with Identifiers - it was added to the story after recent decision made by PO.

PBobylev (changes requested):

src/main/resources/profiles/authority-family.json
SKarolFolio: Done

PBobylev (approved):

